PR #19, opened 2019-09-02
Draft vulnerability schema extension

kakumara:
Adding a draft software vulnerability schema for discussion and getting feedback.

Note: the descriptions are not final and to be updated later on, upon structure finalization.

References: https://github.com/jeremylong/DependencyCheck/blob/master/core/src/main/resources/schema/dependency-check.2.2.xsd

### DIFF
--- a/schema/ext/vulnerability-1.0.xsd
+++ b/schema/ext/vulnerability-1.0.xsd
@@ -119,7 +119,7 @@
                 </xs:complexType>
             </xs:element>
             <xs:element name="severity" type="vuln:severityType" minOccurs="0" maxOccurs="1"/>
-            <xs:element name="source" type="vuln:scoreSourceType" minOccurs="0" maxOccurs="1"/>
+            <xs:element name="method" type="vuln:scoreSourceType" minOccurs="0" maxOccurs="1"/>
             <xs:element name="vector" type="xs:string" minOccurs="0" maxOccurs="1">
                 <xs:annotation>
                     <xs:documentation xml:lang="en">
@@ -173,15 +173,15 @@
                     </xs:attribute>
                 </xs:complexType>
             </xs:element>
-            <xs:element name="vulnerabilityScore" minOccurs="0" maxOccurs="1">
+            <xs:element name="ratings" minOccurs="0" maxOccurs="1">
                 <xs:annotation>
                     <xs:documentation xml:lang="en">
-                        List of the vulnerability score(s) as defined by various risk rating methodologies.
+                        List of the vulnerability ratings as defined by various risk rating methodologies.
                     </xs:documentation>
                 </xs:annotation>
                 <xs:complexType>
                     <xs:sequence>
-                        <xs:element name="score" type="vuln:scoreType" minOccurs="0" maxOccurs="unbounded"/>
+                        <xs:element name="rating" type="vuln:scoreType" minOccurs="0" maxOccurs="unbounded"/>
                     </xs:sequence>
                 </xs:complexType>
             </xs:element>
--- a/schema/ext/vulnerability-1.0.xsd
+++ b/schema/ext/vulnerability-1.0.xsd
@@ -1,0 +1,148 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
+           xmlns:vc="http://www.w3.org/2007/XMLSchema-versioning"
+           xmlns:dg="http://cyclonedx.org/schema/ext/vulnerability/1.0"
+           elementFormDefault="qualified"
+           targetNamespace="http://cyclonedx.org/schema/ext/vulnerability/1.0"
+           vc:minVersion="1.0"
+           vc:maxVersion="1.1"
+           version="1.0.0-SNAPSHOT">
+
+    <xs:complexType name="cvssV2">
+        <xs:annotation>
+            <xs:documentation>
+                Defines the structure of a CVSS v2 vulnerability
+                https://www.first.org/cvss/v2/guide
+            </xs:documentation>
+        </xs:annotation>
+        <xs:sequence>
+            <xs:element name="score" type="xs:decimal" minOccurs="1" maxOccurs="1"/>
+            <xs:element name="attackVector" type="xs:string" minOccurs="1" maxOccurs="1"/>
+            <xs:element name="accessComplexity" type="xs:string" minOccurs="1" maxOccurs="1"/>
+            <xs:element name="authentication" type="xs:string" minOccurs="1" maxOccurs="1"/>
+            <xs:element name="confidentialImpact" type="xs:string" minOccurs="1" maxOccurs="1"/>
+            <xs:element name="integrityImpact" type="xs:string" minOccurs="1" maxOccurs="1"/>
+            <xs:element name="availabilityImpact" type="xs:string" minOccurs="1" maxOccurs="1"/>
+            <xs:element name="severity" type="xs:string" minOccurs="1" maxOccurs="1"/>
+        </xs:sequence>
+    </xs:complexType>
+
+    <xs:complexType name="cvssV3">
+        <xs:annotation>
+            <xs:documentation>
+                Defines the structure of a CVSS v3 vulnerability
+                https://www.first.org/cvss/v3.1/specification-document
+            </xs:documentation>
+        </xs:annotation>
+        <xs:sequence>
+            <xs:element name="baseScore" type="xs:decimal" minOccurs="1" maxOccurs="1"/>
+            <xs:element name="attackVector" type="xs:string" minOccurs="1" maxOccurs="1"/>
+            <xs:element name="attackComplexity" type="xs:string" minOccurs="1" maxOccurs="1"/>
+            <xs:element name="privilegesRequired" type="xs:string" minOccurs="1" maxOccurs="1"/>
+            <xs:element name="userInteraction" type="xs:string" minOccurs="1" maxOccurs="1"/>
+            <xs:element name="scope" type="xs:string" minOccurs="1" maxOccurs="1"/>
+            <xs:element name="confidentialityImpact" type="xs:string" minOccurs="1" maxOccurs="1"/>
+            <xs:element name="integrityImpact" type="xs:string" minOccurs="1" maxOccurs="1"/>
+            <xs:element name="availabilityImpact" type="xs:string" minOccurs="1" maxOccurs="1"/>
+            <xs:element name="baseSeverity" type="xs:string" minOccurs="1" maxOccurs="1"/>
+        </xs:sequence>
+    </xs:complexType>
+
+    <xs:complexType name="vulnerability">
+        <xs:annotation>
+            <xs:documentation>Defines the structure of a vulnerability</xs:documentation>
+        </xs:annotation>
+        <xs:sequence>
+            <xs:element name="id" type="xs:normalizedString" minOccurs="1" maxOccurs="1">
+                <xs:annotation>
+                    <xs:documentation>The id of the vulnerability</xs:documentation>
+                </xs:annotation>
+            </xs:element>
+            <xs:element name="source" minOccurs="0" maxOccurs="1">
+                <xs:complexType>
+                    <xs:annotation>
+                        <xs:documentation>The source of the vulnerability</xs:documentation>
+                    </xs:annotation>
+                    <xs:sequence>
+                        <xs:element name="url" type="xs:string" minOccurs="0" maxOccurs="1">
+                            <xs:annotation>
+                                <xs:documentation>The url of the source</xs:documentation>
+                            </xs:annotation>
+                        </xs:element>
+                    </xs:sequence>
+                    <xs:attribute name="name" type="xs:normalizedString" use="required">
+                        <xs:annotation>
+                            <xs:documentation>The name of the source</xs:documentation>
+                        </xs:annotation>
+                    </xs:attribute>
+                </xs:complexType>
+            </xs:element>
+            <xs:element name="cvssV2" type="dg:cvssV2" minOccurs="0" maxOccurs="1"/>
+            <xs:element name="cvssV3" type="dg:cvssV3" minOccurs="0" maxOccurs="1"/>
+            <xs:element name="cwes" minOccurs="0" maxOccurs="1">
+                <xs:complexType>
+                    <xs:sequence minOccurs="0" maxOccurs="unbounded">
+                        <xs:element name="cwe" type="xs:string" minOccurs="0" maxOccurs="1"/>
+                    </xs:sequence>
+                </xs:complexType>
+            </xs:element>
+            <xs:element name="recommendations" minOccurs="0" maxOccurs="1">
+                <xs:complexType>
+                    <xs:annotation>
+                        <xs:documentation>The recommendations for the vulnerability</xs:documentation>
+                    </xs:annotation>
+                    <xs:sequence minOccurs="0" maxOccurs="unbounded">
+                        <xs:element name="recommendation" type="xs:string"/>
+                    </xs:sequence>
+                </xs:complexType>
+            </xs:element>
+            <xs:element name="description" minOccurs="0" maxOccurs="1">
+                <xs:complexType>
+                    <xs:annotation>
+                        <xs:documentation>The description of the vulnerability</xs:documentation>
+                    </xs:annotation>
+                    <xs:sequence>
+                        <xs:element name="explanation" type="xs:string" minOccurs="0" maxOccurs="1">
+                            <xs:annotation>
+                                <xs:documentation>An explanation of the vulnerability</xs:documentation>
+                            </xs:annotation>
+                        </xs:element>
+                        <xs:element name="detection" type="xs:string" minOccurs="0" maxOccurs="1">
+                            <xs:annotation>
+                                <xs:documentation>The detection of the vulnerability</xs:documentation>
+                            </xs:annotation>
+                        </xs:element>
+                    </xs:sequence>
+                </xs:complexType>
+            </xs:element>
+            <xs:element name="causes" minOccurs="0" maxOccurs="1">
+                <xs:complexType>
+                    <xs:annotation>
+                        <xs:documentation>The causes of the vulnerability</xs:documentation>
+                    </xs:annotation>
+                    <xs:sequence minOccurs="0" maxOccurs="unbounded">
+                        <xs:element name="cause" type="xs:string"/>
+                    </xs:sequence>
+                </xs:complexType>
+            </xs:element>
+            <xs:element name="advisories" minOccurs="0" maxOccurs="1">
+                <xs:complexType>
+                    <xs:annotation>
+                        <xs:documentation>The advisories for the vulnerability</xs:documentation>
+                    </xs:annotation>
+                    <xs:sequence minOccurs="0" maxOccurs="unbounded">
+                        <xs:element name="advisory" type="xs:string"/>
+                    </xs:sequence>
+                </xs:complexType>
+            </xs:element>
+        </xs:sequence>
+    </xs:complexType>
+
+    <xs:element name="vulnerabilities">
+        <xs:complexType>
+            <xs:sequence minOccurs="0" maxOccurs="unbounded">
+                <xs:element name="vulnerability" type="dg:vulnerability"/>
+            </xs:sequence>
+        </xs:complexType>
+    </xs:element>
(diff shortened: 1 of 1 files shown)


schema/ext/vulnerability-1.0.xsd
```diff
@@ -0,0 +1,148 @@
...
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
+           xmlns:vc="http://www.w3.org/2007/XMLSchema-versioning"
+           xmlns:dg="http://cyclonedx.org/schema/ext/vulnerability/1.0"
+           elementFormDefault="qualified"
+           targetNamespace="http://cyclonedx.org/schema/ext/vulnerability/1.0"
+           vc:minVersion="1.0"
+           vc:maxVersion="1.1"
+           version="1.0.0-SNAPSHOT">
+
+    <xs:complexType name="cvssV2">
```
stevespringett: For vulnerabilities that reside in a public source (e.g. NVD), the base score attributes are generally good enough. However, without temporal or environmental attributes, CVSS would not allow an organization to fine tune the severity based on their environment.

schema/ext/vulnerability-1.0.xsd
```diff
@@ -0,0 +1,148 @@
...
+            <xs:element name="accessComplexity" type="xs:string" minOccurs="1" maxOccurs="1"/>
+            <xs:element name="authentication" type="xs:string" minOccurs="1" maxOccurs="1"/>
+            <xs:element name="confidentialImpact" type="xs:string" minOccurs="1" maxOccurs="1"/>
+            <xs:element name="integrityImpact" type="xs:string" minOccurs="1" maxOccurs="1"/>
+            <xs:element name="availabilityImpact" type="xs:string" minOccurs="1" maxOccurs="1"/>
+            <xs:element name="severity" type="xs:string" minOccurs="1" maxOccurs="1"/>
+        </xs:sequence>
+    </xs:complexType>
+
+    <xs:complexType name="cvssV3">
```
stevespringett: For vulnerabilities that reside in a public source (e.g. NVD), the base score attributes are generally good enough. However, without temporal or environmental attributes, CVSS would not allow an organization to fine tune the severity based on their environment.

schema/ext/vulnerability-1.0.xsd
```diff
@@ -0,0 +1,148 @@
...
+                        </xs:element>
+                    </xs:sequence>
+                    <xs:attribute name="name" type="xs:normalizedString" use="required">
+                        <xs:annotation>
+                            <xs:documentation>The name of the source</xs:documentation>
+                        </xs:annotation>
+                    </xs:attribute>
+                </xs:complexType>
+            </xs:element>
+            <xs:element name="cvssV2" type="dg:cvssV2" minOccurs="0" maxOccurs="1"/>
```
stevespringett: Not all public vulnerabilities use CVSS. Some (e.g. NPM Advisories) use strings (Critical, High, Moderate, Low) to represent severity. Organizations that use BOMs within their environment typically will not use CVSS, rather OWASP Risk Rating or a similar methodology.

schema/ext/vulnerability-1.0.xsd
```diff
@@ -0,0 +1,148 @@
...
+            <xs:element name="cvssV3" type="dg:cvssV3" minOccurs="0" maxOccurs="1"/>
```
stevespringett: Same for CVSSv3

stevespringett: I would recommend using vectors instead so that individual CVSSv2/v3/v3.1 will not have to be used. Vectors could also be used for OWASP Risk Rating.

schema/ext/vulnerability-1.0.xsd
```diff
@@ -0,0 +1,148 @@
...
+                    </xs:annotation>
+                    <xs:sequence>
+                        <xs:element name="explanation" type="xs:string" minOccurs="0" maxOccurs="1">
+                            <xs:annotation>
+                                <xs:documentation>An explanation of the vulnerability</xs:documentation>
+                            </xs:annotation>
+                        </xs:element>
+                        <xs:element name="detection" type="xs:string" minOccurs="0" maxOccurs="1">
+                            <xs:annotation>
+                                <xs:documentation>The detection of the vulnerability</xs:documentation>
```
stevespringett: This needs more clarification. What is detection?

schema/ext/vulnerability-1.0.xsd
```diff
@@ -0,0 +1,148 @@
...
+                            </xs:annotation>
+                        </xs:element>
+                    </xs:sequence>
+                </xs:complexType>
+            </xs:element>
+            <xs:element name="causes" minOccurs="0" maxOccurs="1">
```
stevespringett: Wouldn't cause be associated to a CWE?

schema/ext/vulnerability-1.0.xsd
kakumara: > For vulnerabilities that reside in a public source (e.g. NVD), the base score attributes are generally good enough. However, without temporal or environmental attributes, CVSS would not allow an organization to fine tune the severity based on their environment.

Taking above in to consideration and  https://github.com/CycloneDX/specification/pull/19#discussion_r320037371 restructured the scoreType to be generic that can support multiple risk rating standards. Most of the elements have been made optional.

schema/ext/vulnerability-1.0.xsd
kakumara: > For vulnerabilities that reside in a public source (e.g. NVD), the base score attributes are generally good enough. However, without temporal or environmental attributes, CVSS would not allow an organization to fine tune the severity based on their environment.

👍 Taking above in to consideration and  https://github.com/CycloneDX/specification/pull/19#discussion_r320037371 restructured the scoreType to be generic that can support multiple risk rating standards. Most of the elements have been made optional.

schema/ext/vulnerability-1.0.xsd
kakumara: > Not all public vulnerabilities use CVSS. Some (e.g. NPM Advisories) use strings (Critical, High, Moderate, Low) to represent severity. Organizations that use BOMs within their environment typically will not use CVSS, rather OWASP Risk Rating or a similar methodology.

👍 restructured the scoreType to be generic that can support multiple risk rating methodologies.

schema/ext/vulnerability-1.0.xsd
kakumara: > I would recommend using vectors instead so that individual CVSSv2/v3/v3.1 will not have to be used. Vectors could also be used for OWASP Risk Rating.

Removed CVSSv2/CVSSv3 and replaced with a generic risk score structure. CVSSv2

schema/ext/vulnerability-1.0.xsd
kakumara: > This needs more clarification. What is detection?

This is used as an attribute to describe vulnerabilities in sonatype products. Removed it considering the generic nature of this schema.  efa5e75

schema/ext/vulnerability-1.0.xsd
kakumara: > Wouldn't cause be associated to a CWE?

Not as per our research and how its been adopted in sonatype products. Similar to above removed it considering the generic nature of this schema. efa5e75

schema/ext/vulnerability-1.0.xsd
```diff
@@ -0,0 +1,132 @@
...
+                    </xs:annotation>
+                    <xs:sequence minOccurs="0" maxOccurs="unbounded">
+                        <xs:element name="advisory" type="xs:string"/>
+                    </xs:sequence>
+                </xs:complexType>
+            </xs:element>
+        </xs:sequence>
+    </xs:complexType>
+
+    <xs:element name="vulnerabilities">
```
stevespringett: Is the intended use of the vulnerabilities node to reside inside each component?

```xml
<bom>
  <components>
    <component>
      <dg:vulnerabilities>
        <dg:vulnerability>
        </dg:vulnerability>
      </dg:vulnerabilities>
    </component>
  </components>
</bom>
```

Whatever the intended location, guidance in the form of documentation would be necessary

ja98200wilcox: > Is the intended use of the vulnerabilities node to reside inside each component?

That is the intended location to be inside a component element.  Seems like a natural fit.  Would you rather have it somewhere else?

stevespringett: Makes perfect sense to me as well, but the element wasn't annotated with documentation that provided any guidance. So although I assumed this to be the case, the specification did not explicitly state that.

ja98200wilcox: > guidance in the form of documentation would be necessary

Will add appropriate documentation elements that provide guidance

schema/ext/vulnerability-1.0.xsd
```diff
@@ -10,6 +10,7 @@
 
     <xs:simpleType name="severityType" final="restriction">
         <xs:restriction base="xs:normalizedString">
+            <xs:enumeration value="None"/>
```
stevespringett: I think 'Unknown' may be a better term. 'None' could be confused to mean a INFO level warning rather than a vulnerability of unknown severity.

schema/ext/vulnerability-1.0.xsd
```diff
@@ -0,0 +1,268 @@
...
+                        </xs:element>
+                    </xs:sequence>
+                </xs:complexType>
+            </xs:element>
+            <xs:element name="severity" type="dg:severityType" minOccurs="0" maxOccurs="1"/>
+            <xs:element name="source" type="dg:scoreSourceType" minOccurs="0" maxOccurs="1"/>
+            <xs:element name="vector" type="xs:string" minOccurs="0" maxOccurs="1">
+                <xs:annotation>
+                    <xs:documentation xml:lang="en">
+                        textual representation of the metric values used to score the vulnerability
```
mr-zepol: minor: change `textual` for `Textual`, looks better, since the other documentation starts with capital letter

schema/ext/vulnerability-1.0.xsd
```diff
@@ -0,0 +1,268 @@
...
+                                The name of the source. For example "National Vulnerability Database"
+                            </xs:documentation>
+                        </xs:annotation>
+                    </xs:attribute>
+                </xs:complexType>
+            </xs:element>
+            <xs:element name="vulnerabilityScore" minOccurs="0" maxOccurs="1">
+                <xs:annotation>
+                    <xs:documentation xml:lang="en">
+                        A vulnerability can have zero or more scores from different risk scoring methodologies.
```
mr-zepol: minor: `more scores` implies that there could be many of them, but `maxOccurs` is only one, might be confusing

schema/ext/vulnerability-1.0.xsd
kakumara: > minor: change `textual` for `Textual`, looks better, since the other documentation starts with capital letter

👍 5c75f6f

schema/ext/vulnerability-1.0.xsd
kakumara: > minor: `more scores` implies that there could be many of them, but `maxOccurs` is only one, might be confusing

I've reworded it slightly differently 5c75f6f

schema/ext/vulnerability-1.0.xsd
ja98200wilcox: > I think 'Unknown' may be a better term. 'None' could be confused to mean a INFO level warning rather than a vulnerability of unknown severity.


'None' came from the cvss ratings.  My understanding is 'None' means there’s next to no threat.  'Unknown' would be the score hasn’t been calculated.  

Screen Shot 2019-09-06 at 2 22 32 PM

ja98200wilcox:
@stevespringett How does the vulnerability schema get tied to the SBOM schema?  The vulnerability schema documentation is the tie?

stevespringett:
@ja98200wilcox 
`bom-ref` might be an option as it mandates a unique identifier for each component. Then for each vulnerability, you could potentially refer to each component via its `bom-ref`.

I would prefer this schema to be flexible enough to be used inside and outside a component. Look at externalResources as an example. It supports both a bom-level representation and inside the component as well. If the vulnerability schema is used outside of a component, it will need to use the `bom-ref`.

The two use-cases that I see this schema being used for are:
* Specifying zero or more vulnerabilities for each component
* Specifying zero or more vulnerabilities for the asset/application that a bom represents

schema/ext/vulnerability-1.0.xsd
kakumara: >'None' came from the cvss ratings. My understanding is 'None' means there’s next to no threat. 'Unknown' would be the score hasn’t been calculated.

I've added `Unknown` the list of options as well. c3af551

ja98200wilcox:
> bom-ref might be an option as it mandates a unique identifier for each component. 

I see the dependency-graph-1.0.xsd does that.  Thank you.  Didn't see that tie in before.

ja98200wilcox:
> bom-ref might be an option as it mandates a unique identifier for each component.

Added bom-ref to vulnerability 
`<vulnerability ref="string">`

ja98200wilcox:
Hi @stevespringett, anything else you like changed?  What are the next steps you like us to do?

stevespringett:
I will review this weekend and provide feedback. Once released, the extension will have a page similar to https://cyclonedx.org/ext/dependency-graph/

schema/ext/vulnerability-1.0.xsd
```diff
@@ -0,0 +1,274 @@
...
+            </xs:enumeration>
+            <xs:enumeration value="CVSSv3">
+                <xs:annotation>
+                    <xs:documentation xml:lang="en">
+                        The rating is based on CVSS v3 standard
+                        https://www.first.org/cvss/v3.1/specification-document
+                    </xs:documentation>
+                </xs:annotation>
+            </xs:enumeration>
+            <xs:enumeration value="NPM advisory">
```
stevespringett: I think NPM advisory would quality as 'Other'. This should likely be removed

schema/ext/vulnerability-1.0.xsd
```diff
@@ -0,0 +1,274 @@
...
+                        <xs:element name="url" minOccurs="0" maxOccurs="1">
+                            <xs:annotation>
+                                <xs:documentation xml:lang="en">
+                                    The url of the vulnerability documentation as provided by the source
+                                    For example https://nvd.nist.gov/vuln/detail/CVE-2019-15842
+                                </xs:documentation>
+                            </xs:annotation>
+                            <xs:simpleType>
+                                <xs:restriction base="xs:string">
+                                    <xs:pattern value="(^(([^:/?#]+):)?(//([^/?#]*))?([^?#]*)(\?([^#]*))?(#(.*))?)"/>
```
stevespringett: In the doc above, https://nvd.nist.gov/vuln/detail/CVE-2019-15842 does nit validate against the pattern specified. Regex in XML Schema is very limited - not all features are supported. You may have to work within those limitations, correct the expression, or use anyURI instead. URI obviously not as strict as URL.

schema/ext/vulnerability-1.0.xsd
kakumara: > I think NPM advisory would quality as 'Other'. This should likely be removed

Fair enough 👍 35bb842

schema/ext/vulnerability-1.0.xsd
kakumara: > In the doc above, https://nvd.nist.gov/vuln/detail/CVE-2019-15842 does not validate against the pattern specified. Regex in XML Schema is very limited - not all features are supported. You may have to work within those limitations, correct the expression, or use anyURI instead. URI obviously not as strict as URL.

Yup, `anyURI` seems a better fit. 35bb842

stevespringett:
I think we're really close. I see a few minor issues when using the schema:

```xml
<v:vulnerabilityScore>
    <v:score>
        <v:score>
            <v:base>9.8</v:base>
            <v:impact>5.9</v:impact>
            <v:exploitability>3.0</v:exploitability>
        </v:score>
        <v:severity>Critical</v:severity>
        <v:source>CVSSv3</v:source>
        <v:vector>AV:N/AC:L/PR:N/UI:N/S:U/C:H/I:H/A:H</v:vector>
    </v:score>
</v:vulnerabilityScore>
```

The `<vulnerabilityScore>` should be renamed to a plural. 'ratings' maybe?

Inside `<ratings>` should be a `<rating>` and inside that should be the `<score>`. Having two `<score>` nodes with different meanings is not acceptable.

Proposed changes:
```xml
<v:ratings>
    <v:rating>
        <v:score>
            <v:base>9.8</v:base>
            <v:impact>5.9</v:impact>
            <v:exploitability>3.0</v:exploitability>
        </v:score>
        <v:severity>Critical</v:severity>
        <v:source>CVSSv3</v:source>
        <v:vector>AV:N/AC:L/PR:N/UI:N/S:U/C:H/I:H/A:H</v:vector>
    </v:rating>
</v:ratings>
```

If you have a proposal other than 'ratings' that also works or is more descriptive of the nodes, please let me know.

The final thing that needs correction is the score decimal restrictions. Currently, any decimal is allowed. So -999, 15.8, and 5.99999 are all allowed. This should be restricted to 0.0 - 10.0 and be limited to a single decimal point.

stevespringett:
One final point on the above. The `<v:source>CVSSv3</v:source>` should likely be changed to `<v:method>CVSSv3</v:method>`. So we will be describing the method used in risk rating the vulnerability. To me, that makes more sense.

kakumara:
> The `<vulnerabilityScore>` should be renamed to a plural. 'ratings' maybe?
> 
> Inside `<ratings>` should be a `<rating>` and inside that should be the `<score>`. Having two `<score>` nodes with different meanings is not acceptable.

Agree and makes sense. 44a9bfa

kakumara:
> One final point on the above. The `<v:source>CVSSv3</v:source>` should likely be changed to `<v:method>CVSSv3</v:method>`. So we will be describing the method used in risk rating the vulnerability. To me, that makes more sense.

Agree and makes sense. 44a9bfa

ja98200wilcox:
> Once released, the extension will have a page similar to https://cyclonedx.org/ext/dependency-graph/

Does Sonatype create this or @stevespringett?

Also, I would like to thank you for reviewing our PR.

stevespringett:
> Does Sonatype create this or @

This is something I’ll  do and if you’d like the page to state something specific or if Sonatype would like to be ack on the page, let me know.

> Also, I would like to thank you for reviewing our PR.

Of course. I think the only thing remaining is the decimal restrictions. Other than that, does the PR satisfy your original requirements? Any foreseeable changes or is this ready for release?

kakumara:
>The final thing that needs correction is the score decimal restrictions. Currently, any decimal is allowed. So -999, 15.8, and 5.99999 are all allowed. This should be restricted to 0.0 - 10.0 and be limited to a single decimal point.

👍 restriction added 57d27d8

stevespringett:
How will the schema be licensed? CycloneDX itself is Apache 2.0.

schema/ext/vulnerability-1.0.xsd
```diff
@@ -0,0 +1,275 @@
...
+            </xs:enumeration>
+            <xs:enumeration value="Open FAIR">
+                <xs:annotation>
+                    <xs:documentation xml:lang="en">
+                        The rating is based on Open FAIR specification
+                        http://www.opengroup.org/subjectareas/security/risk
+                    </xs:documentation>
+                </xs:annotation>
+            </xs:enumeration>
+            <xs:enumeration value="PCI DSS">
```
stevespringett: I looked at the PCI DSS doc and it doesn't specifically state a method for evaluating risk. Rather, it provides an example of how to do so. The only requirement appears to be that it's done. But I'm unclear if PCI DSS should actually be a valid risk rating method or not. Comments?

ja98200wilcox:
> How will the schema be licensed? CycloneDX itself is Apache 2.0.

Let's use the same license you use for CycloneDX.

schema/ext/vulnerability-1.0.xsd
kakumara: > I looked at the PCI DSS doc and it doesn't specifically state a method for evaluating risk. Rather, it provides an example of how to do so. The only requirement appears to be that it's done. But I'm unclear if PCI DSS should actually be a valid risk rating method or not. Comments?

Think you are right. 👍  This came as an option in our initial research. But based on above and internal discussion this doesn't look like it can be placed as a risk rating method.   b2e2b00

stevespringett:
I'll add the license and publish this weekend.

Issue #38, opened 2020-10-13
Promote and enhance vulnerability schema


garethr:
Thanks for creating the JSON Schema variant of the vulnerability extension. This prompted me to take a run at describing some real world data using the format. Here's the full example https://gist.github.com/garethr/7dcc9d6ef4e7cc497e018dc279c00123 (bias warning, I work for Snyk, so I used Snyk in the comparison. I don't think any of the following is Snyk specific though, more about the general complexity of the domain and general usage of the output.)

Here are my observations from that experiment. Happy to discuss these here in one long thread or break out elsewhere if easier. Posting partly to discover the best way to discuss and appetite for addressing issues agreed after discussion.

1. No fields are required. This makes the life of consumers harder, should there be a minimum set of fields? (`id`, `description`?) 
2. The score is always base/impact/exploitability, and always 0-10, so only compatible with CVSS
3. A vulnerability refers to a BOM ref, which I think means if multiple components have the same vulnerability then it needs to be included multiple times rather than via reference?
4. No provision for multiple vulnerability identifiers for the same vulnerability, ie. CVE-2009-5155 is also SNYK-DEBIAN9-GLIBC-338103
5. Although method on ratings can be set to "Other", there is no way of specifying what method/source the information is coming from. So you could have multiple other sources, but no way for the user to trace those back to that source based on data
6. In the case of multiple ratings with the same method, but different scores/vectors, what are the heuristics? Is this valid? Should the spec prohibit this?
7. Additional properties are allowed as per the JSON Schema. Is this purposeful? What is the expected parser behaviour for attributes outside the schema? Should there be a formal extension mechanism?
8. The description is a string. Are there are rules here, ie. SARIF has plain text and markdown properties explicitly, rather than leaving up to the producer
9. Are advisories intended to always be URLs, or can they be arbitrary strings? The schema says the latter, although all the examples are the former. This would impact someone writing a parser.

ie. Should advisories be defined as:

```yaml
{
    "type": "string",
    "format": "uri",
    "pattern": "^(https?|http?)://"
}
```

Hopefully the above is useful feedback, rather than being seen as too much criticism. Some might be me being new to the spec and missing things.

I'd be happy to help, and to discuss further if useful.

coderpatros:
Thanks for the feedback @garethr, it is very much appreciated!

stevespringett:
[...]

PR #44, opened 2021-01-01
Proposed evolution of the vulnerability schema for discussion

garethr:
WIP based on discussions in https://github.com/CycloneDX/specification/issues/38

* More prescriptive definition for advisories
* Support multiple sources for each vulnerability
* Allow ratings to have a named source

I've started with the JSON Schema just to iterate on ideas and get feedback. Would update examples and XSD and other parts before moving out of draft.

### DIFF
--- a/schema/ext/vulnerability-2.0-SNAPSHOT.schema.json
+++ b/schema/ext/vulnerability-2.0-SNAPSHOT.schema.json
@@ -73,41 +73,54 @@
       ]
     },
     "score": {
-      "type": "object",
-      "title": "Score",
-      "description": "Defines the numerical risk score of a vulnerability",
-      "properties": {
-        "base": {
-          "type": "number",
-          "title": "Base Score",
-          "description": "The base score of the security vulnerability (Refer CVSS standard for example)",
-          "multipleOf": 0.1,
-          "examples": [
-            2.9,
-            7.2
-          ]
+      "oneOf": [
+        {
+          "type": "number"
         },
-        "impact": {
-          "type": "number",
-          "title": "Impact Score",
-          "description": "The impact subscore of the security vulnerability (Refer CVSS standard for example)",
-          "multipleOf": 0.1,
-          "examples": [
-            2.9,
-            7.2
-          ]
-        },
-        "exploitability": {
-          "type": "number",
-          "title": "Exploitability Score",
-          "description": "The exploitability subscore of the security vulnerability (Refer CVSS standard for example)",
-          "multipleOf": 0.1,
-          "examples": [
-            2.9,
-            7.2
-          ]
+        {
+          "type": "object",
+          "title": "Score",
+          "description": "Defines the numerical risk score of a vulnerability",
+          "properties": {
+            "base": {
+              "type": "number",
+              "title": "Base Score",
+              "description": "The base score of the security vulnerability (Refer CVSS standard for example)",
+              "multipleOf": 0.1,
+              "minimum": 0,
+              "maximum": 10,
+              "examples": [
+                2.9,
+                7.2
+              ]
+            },
+            "impact": {
+              "type": "number",
+              "title": "Impact Score",
+              "description": "The impact subscore of the security vulnerability (Refer CVSS standard for example)",
+              "multipleOf": 0.1,
+              "minimum": 0,
+              "maximum": 10,
+              "examples": [
+                2.9,
+                7.2
+              ]
+            },
+            "exploitability": {
+              "type": "number",
+              "title": "Exploitability Score",
+              "description": "The exploitability subscore of the security vulnerability (Refer CVSS standard for example)",
+              "multipleOf": 0.1,
+              "minimum": 0,
+              "maximum": 10,
+              "examples": [
+                2.9,
+                7.2
+              ]
+            }
+          }
         }
-      }
+      ]
     },
     "rating": {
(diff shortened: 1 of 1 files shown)


coderpatros:
@garethr Do you already have an example JSON file for this you can share?

garethr:
@coderpatros Yes, I'd linked in the original thread and not here. Here's a simple example https://gist.github.com/garethr/b069d9bf84ca80635cd506e74c8e2247

coderpatros:
@stevespringett do you see any blockers here?

stevespringett:
No, I don't see any blockers here. I think there's some good ideas in here, but would like to see side-by-side examples of what can be done with 1.0 vs 2.0.  Looking at the schema, I see there's a bit more flexibility in some regards, like having multiple sources for a vulnerability.

garethr:
+1 on more examples. Do you have a batch of v1.0 examples? I'd be happy to port those for comparison.

I think the other question is, are there other backwards incompatible changes that would be good to make now for a major?

schema/ext/vulnerability-2.0-SNAPSHOT.schema.json
```diff
@@ -0,0 +1,235 @@
...
+    },
+    "score": {
+      "oneOf": [
+        {
+          "type": "number"
+        },
+        {
+          "type": "object",
+          "title": "Score",
+          "description": "Defines the numerical risk score of a vulnerability",
```
pombredanne, line 73: You mentioned in your FOSDEM talk that you wanted to support more than CVSS, but is this not exactly CVSS scoring?

schema/ext/vulnerability-2.0-SNAPSHOT.schema.json
```diff
@@ -0,0 +1,235 @@
...
+          "title": "Source",
+          "description": "The entity responsible for providing the rating",
+          "examples": [
+            "NVD"
+          ]
+        },
+        "vector": {
+          "type": "string",
+          "title": "Vector",
+          "description": "Textual representation of the metric values used to score the vulnerability see attack vector in https://www.first.org/cvss/v3.1/specification-document"
```
pombredanne, line 140: Are you not assuming then that all rating systems use CVSS and a vector?

schema/ext/vulnerability-2.0-SNAPSHOT.schema.json
garethr: See the addition of "oneOf" above and the number type https://github.com/CycloneDX/specification/pull/44/files#diff-da04b5360b00e6b44767824827b36a44cf40f651723db0db12d1e70ea8e353a6R68

This means cores are either a CVSS score, or a number. This might want to be broader still. Collecting examples of simple and complex scores would be useful.

schema/ext/vulnerability-2.0-SNAPSHOT.schema.json
garethr: None of these are required fields, which is a bit of a design decision with CycloneDX I feel based on some of the questions I asked previously. So ratings here can be any of these individually or together.

devkadirselcuk (approved):

stevespringett:
@garethr Due to the extensibility issues with JSON, we will be moving the vulnerability extension into the core spec. This is scheduled for v1.4 which is already in progress. I'd love to get your feedback on it when a proposal is available. I'll take into consideration the design elements you've included in this PR.

schema/ext/vulnerability-2.0-SNAPSHOT.schema.json
```diff
@@ -0,0 +1,235 @@
...
+          "type": "string",
+          "format": "string",
+          "title": "Reference",
+          "description": "References a component by the components bom-ref attribute"
+        },
+        "id": {
+          "type": "string",
+          "title": "ID",
+          "description": "The id of the vulnerability as defined by the risk scoring methodology. For example CVE-2019-15842 (of https://nvd.nist.gov/vuln/detail/CVE-2019-15842)"
+        },
```
stevespringett, line 186: @garethr I'm curious why you have a single ID, but have multiple sources.  Wouldn't each source potentially have its own ID?

CVE-xxxx-xxxx
   aka
SNYK-xxxxxxxxx
   aka
SONATYPE-xxxxxxxx

stevespringett: Also, wouldn't it be possible that each source could potentially have slightly different ratings as well? I'm thinking of VulnDB for example which makes a ton of corrections and enhancements to the NVD and its not uncommon for them to provide both the NVD scores as well as their own.

Are you trying to capture the ability to specify that a vulnerability will be known by different names across different sources?

schema/ext/vulnerability-2.0-SNAPSHOT.schema.json
stevespringett: I'm not sure what CVSSv4 will bring, and even if we'll continue to have base, impact, and exploitability scores. But the interesting thing is, this information is actually redundant. The scoring type is specific and if the vector is also specified, the scores can be recomputed. We can likely simplify this to a single score. This is what I'm planning to do with v1.4

stevespringett:
Closing as vulnerability support has been greatly expanded and incorporated into the core spec - thus available for JSON and Protobuf. See #91

Issue #49, opened 2021-03-12
Add attached text type for components and BOM level metadata


coderpatros:
RFC #46

PR #50, opened 2021-03-12
Add attached text type for components and BOM level metadata

coderpatros:
First cut of #49 

Feedback most welcome. Especially before I go to the trouble of making tests :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+tools/
+.idea/
--- a/schema/bom-1.3-SNAPSHOT.schema.json
+++ b/schema/bom-1.3-SNAPSHOT.schema.json
@@ -116,6 +116,12 @@
           "title": "Supplier",
           "description": " The organization that supplied the component that the BOM describes. The supplier may often be the manufacture, but may also be a distributor or repackager.",
           "$ref": "#/definitions/organizationalEntity"
+        },
+        "attachments": {
+          "type": "array",
+          "title": "Attachments",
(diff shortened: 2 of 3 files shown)


stevespringett:
My only recommendation for this is to include an OPTIONAL `name` or `title` that describes what is being attached.

Example:
```xml
<attachments>
    <attachment name="Installation Instructions" content-type="text/plain" encoding="base64">VGVzdGluZyBUZXN0aW5nIDEtMi0z</attachment>
</attachments>
```

I have no opinion on whether its called `name` or `title` or something else. If there is somewhere else in the schema where we use a similar label, we should strive for consistence where possible.

stevespringett:
What impact will this have on the other fields that rely on attachmentType? I think license and swid rely on this. Would this recommendation break or add confusion to other uses?

coderpatros:
I'm thinking maybe leave attached text as is. And maybe something like this? With the contents element being the existing attachedTextType.

```
<attachments>
    <attachment>
        <name>Installation Instructions</name>
        <description>extended description</description>
        <contents content-type="text/plain" encoding="base64">VGVzdGluZyBUZXN0aW5nIDEtMi0z</contents>
    </attachment>
</attachments>
```

stevespringett:
Good suggestion as it doesn't impact existing uses, but gives additional context to this use case.

coderpatros:
No test data yet, but reviews welcome.

Issue #42, opened 2020-11-01
RFC: Add per component and per BOM key/value pairs 


stevespringett:
There are times when adopters and implementors of CycloneDX need to specify data that falls outside of what the spec provides. Vendor specific schema extensions are one way to accomplish that, however, it also leads to vendor lock-in.

Therefore, it is desirable to support per component and per bom key/value stores.

For example:
```xml
<component type="library">
    <name>Acme Component</name>
    <version>1.0.0</version>
    <properties>
        <property name="myprop" value="myvalue"/>
    </properties>
</component>
```

This allows a certain degree of customization without having to use vendor-specific extensions. Support for properties inside of other objects (e.g. supplier and license) may also be desirable.

coderpatros:
I really like this addition.

It supports use cases like internal (semi-)automated workflows. I propose this is also included at the BOM level.

stevespringett:
@wrgoff This is still being designed. Once complete, if the RFC is approved, then it will be incorporated into the next version of the CycloneDX specification (v1.3). Historically, there has been a new release of CycloneDX in Q1/Q2 every year for the past three years.

brianf:
I'm +1 to this addition as well.

stevespringett:
Reopening. For XML, I think it makes sense to implement the value as a child element rather than an attribute.

The original proposal specified:
```xml
<properties>
    <property name="myprop" value="myvalue"/>
</properties>
```

However, I think it might be better to support it like this:
```xml
<properties>
    <property name="myprop">myvalue</property>
</properties>
```

By doing it this way, adopters could specify CDATA in value to minimize escaping.

Thoughts? Does it matter?

The way its implemented in JSON would remain the same.

coderpatros:
I think the second option is better.

There's also inconsistency between the JSON and XML schema. `name` for XML and `key` for JSON.

```
<properties>
    <property name="myprop">myvalue</property>
</properties>
```

```
"properties": [
  {
    "key": "Foo",
    "value": "Bar"
  }
]
```

coderpatros:
Actually, I much prefer name. When I hear key value I think the key would be unique.

stevespringett:
You bring up a good point @coderpatros. While implementing support for this I originally started with a hashmap, then realized it would not work since duplicates are allowed in the spec. I'll change it to name/value for both and ensure the value in XML is an element rather than an attribute.

stevespringett:
Ok, the changes have been made. The following syntax is now used:

[...]

PR #55, opened 2021-04-03
Added support for key/value store (properties)

stevespringett:
This PR implements #42 for the BOM itself (metadata) and for an individual component and service.


coderpatros (approved):
LGTM

Just wondering if the tests should include multiple properties with the same key and different values? Just in case anyone (me) is using core spec tests for implementations :)

PR #58, opened 2021-04-11
manufacture grammar fix

bradh:
Follow-on from #57, which on balance seems like a bad idea.

This just fixes the descriptive text part for 1.3 schemas, which should be a compatible change.


stevespringett:
Thanks for the PR

PR #59, opened 2021-04-13
Initial implementation for compositions (known unknowns)

stevespringett:
#48


tools/src/test/resources/valid-compositions-1.3.xml
```diff
@@ -0,0 +1,51 @@
...
+                <dependency ref="acme-application-1.0"/>
+            </dependencies>
+        </composition>
+        <composition>
+            <aggregate>unknown</aggregate>
+            <assemblies>
+                <assembly ref="pkg:maven/acme/library@3.0"/>
+            </assemblies>
+        </composition>
+    </compositions>
```
coderpatros, line 50: In this example, is this the correct interpretation?

- acme-application has a complete list of dependencies, and not_specified for assemblies/components
- shaded library assemblies/sub-components are complete (the sub-component is unspecified for dependencies and assemblies)
- acme library has unknown sub-components, and unspecified dependencies

schema/bom-1.3-SNAPSHOT.schema.json
```diff
@@ -77,6 +77,14 @@
...
       "description": "Provides the ability to document dependency relationships."
+    },
+    "compositions": {
+      "$id": "#/properties/compositions",
+      "type": "array",
+      "items": {"$ref": "#/definitions/compositions"},
+      "uniqueItems": true,
+      "title": "Compositions",
+      "description": "Compositions describe constituent parts (including components, services, and dependency relationships) and their completeness."
     }
```
coderpatros, line 88: I'm wondering, given JSON format limitations, if compositions should be an object with an array of items as a property.

stevespringett: What the PR is saying is that compositions is an array of composition objects (I should change this from compositions to composition). The composition is an object, I think we'll put all the composition information in there. I don't know if we would ever put non-composition objects in the compositions array though.

tools/src/test/resources/valid-compositions-1.3.xml
stevespringett: That is absolutely correct.

schema/bom-1.3-SNAPSHOT.schema.json
coderpatros: Yep, what I'm proposing is something like this in the JSON BOM

```
compositions: {
    items: []
```

Just in case we think we might want to add any additional attributes to compositions element in the future.

Just a thought, I don't feel strongly about it at all.

PR #61, opened 2021-04-18
Added support for evidence of licenses and copyrights

stevespringett:
Implements #43 

Components can now contain evidence of licenses and copyright text. If a component is a library/framework, that information will be applicable to the package. If the component is a file, then the license/copyright evidence is specific to the file.


tools/src/test/resources/valid-evidence-1.3.xml
```diff
@@ -0,0 +1,35 @@
...
+                        <id>LGPL-2.1-only</id>
+                        <url>https://opensource.org/licenses/LGPL-2.1</url>
+                    </license>
+                </licenses>
+                <copyright>
+                    <text><![CDATA[Copyright 2012 Google Inc. All Rights Reserved.]]></text>
+                    <text><![CDATA[Copyright (C) 2004,2005 Dave Brosius <dbrosius@users.sourceforge.net>]]></text>
+                    <text><![CDATA[Copyright (C) 2005 William Pugh]]></text>
+                    <text><![CDATA[Copyright (C) 2004,2005 University of Maryland]]></text>
+                </copyright>
```
coderpatros, line 31: Should these be represented more like this? Just thinking if we want to add anything to it in the future.

```
<copyrights>
  <copyright>
    <text>Copyright (c) 2000 Bleep</text>
  <copyright>
  <copyright>
    <text>Copyright (c) 2010 Bloop</text>
  <copyright>
</copyrights>
```

stevespringett: Sure, they can me. It doesn't require any changes to the schema. But I put everything in CDATA because the second entry containing the email address will make the XML invalid without the CDATA.

coderpatros: Sorry, ignore the cdata part, that was unintentional. I meant the xml element hierarchy: compyrights>copyright>text as opposed to copyrights>text

stevespringett: I did it this way in the event we ever need to specify the file, line number, or any other type of data about where the copyright text was found. Not ideal, yes. But I think there may be some future room for improvement here in the event people start adopting it.

coderpatros: I guess that stuff could be attributes on the text element. Yeah, never mind me :)

PR #63, opened 2021-04-21
Specification v1.3

coderpatros:
Tracking updates to v1.3 branch in prep for the release


schema/bom-1.3-SNAPSHOT.proto
```diff
@@ -0,0 +1,391 @@
...
+  string name = 2;
+  // Specifies the optional full text of the attachment
+  AttachedText text = 3;
+  // The URL to the attachment file. If the attachment is a license or BOM, an externalReference should also be specified for completeness.
+  string url = 4;
+}
+
+message Metadata {
+  // The date and time (timestamp) when the document was created.
+  uint64 timestamp = 1;
```
DarthHater, line 261: You can alternatively use `google.protobuf.Timestamp` here, and import it `import "google/protobuf/timestamp.proto";`, 

it's been available since proto3 and might add more flexibility/stronger typing (the `uint64` doesn't force a UNIX timestamp, but encourages it)

schema/bom-1.3-SNAPSHOT.proto
```diff
@@ -0,0 +1,391 @@
+syntax = "proto3";
+package org.cyclonedx.schema.bom._1_3;
+
```
DarthHater, line 3: One thing to consider in general for protobuf is `optional` https://stackoverflow.com/questions/42622015/how-to-define-an-optional-field-in-protobuf-3/62566052#62566052

There are some nice advantages to doing this for an upstream consumer, because code gen can create methods to check if a field exists, etc...

If a field is truly optional, this would be nice for people consuming. It will prevent a lot of null checks, etc... I imagine.

schema/bom-1.3-SNAPSHOT.proto
coderpatros: Thanks @DarthHater. Is that pretty much universally supported across protobuf implementations?

DarthHater: For `proto3` I believe so! Being you specified `proto3` you should be good to go! This stackoverflow post is where I'm gleaning some of that info from, if you want to chase it further: https://stackoverflow.com/a/42922721/338597

DarthHater:
@coderpatros would it be desirable to have an example protocol buffer project? I cooked one up a few months ago when I was looking at them, I can likely open source it and show how to use the spec to create a service, etc... for people?

schema/bom-1.3-SNAPSHOT.proto
coderpatros: Thanks @DarthHater, I’ll do some further work on the proto file in a seperate PR and tag you in it.

coderpatros:
@DarthHater I think an example project would be really good.

DarthHater:
@coderpatros I'll set one up and open source it!

schema/bom-1.3-SNAPSHOT.proto
coderpatros: New version uses `google.protobuf.Timestamp`

schema/bom-1.3-SNAPSHOT.proto
coderpatros: @DarthHater Do you know how `optional` works on the wire and in practice?

I was just reading the documentation here https://developers.google.com/protocol-buffers/docs/proto#optional. I don't really understand the value unless you want to assign some kind of non-default default value.

But, then reading this https://github.com/protocolbuffers/protobuf/blob/v3.12.0/docs/field_presence.md and it seems to do more.

Is it actually storing an extra information as to the presence of a null value? Or just indicating to codegen that it should create helper methods?

coderpatros: Ah, sorry, that first link is to the proto2 documentation.

coderpatros:
External reference hashes have been add to protobuf and tested in .NET implementation

coderpatros:
@stevespringett this test failure might be one of those you mentioned re java core schemas.

stevespringett:
Vote to merge to master

coderpatros:
I fully support merging this in.

PR #65, opened 2021-04-27
Refactor BOM license to make use of license choice type

coderpatros:
Should not be any material change. Just removing duplicate definitions.

### DIFF
--- a/schema/bom-1.3-SNAPSHOT.schema.json
+++ b/schema/bom-1.3-SNAPSHOT.schema.json
@@ -128,30 +128,7 @@
         "licenses": {
           "type": "array",
           "title": "BOM License(s)",
-          "items": {
-            "properties": {
-              "license": {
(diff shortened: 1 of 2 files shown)


stevespringett (approved):
Looks good. Will likely need a test for this if one doesn't exist already.

coderpatros:
We do have a basic valid/invalid test for this. But it only flexes the license ID scenario.

But I figure we already have the other scenarios covered at the component level. So not sure whether or not we should bother with the extras at the metadata level.

Thoughts @stevespringett?

stevespringett:
> So not sure whether or not we should bother with the extras at the metadata level.

Given the license is a defined and reusable type, I do not see the value in creating extensive testing for everywhere a license can be used. IMO, as long as we test that a license can be used in the context of the BOM, and we at least have one set of extensive tests for the type, that should be enough.

PR #66, opened 2021-04-27
Tracking updates to protobuf format for feedback

coderpatros:


coderpatros:
I added a spec_version field to help when processing a BOM protobuf message. Obviously not of much value right now. But will be important, IMHO, when future versions are released.

Also changed to `google.protobuf.timestamp` for timestamps and explicit oneof declarations for related license element choices.

@DarthHater, I started adding `optional` to fields but ended up with this message when validating protobuf message payloads

```
Explicit 'optional' labels are disallowed in the Proto3 syntax. To define 'optional' fields in Proto3, simply remove the 'optional' label, as fields are 'optional' by default.
```

I have read about optional being opt in behavior. Do you know if this is one of those non-default things that most people do? From my reading it does seem like optional could be beneficial.

coderpatros:
Hi @prabhu, just tagging you in case you, or one of the team, want to provide feedback on this.

coderpatros:
Merging this into v1.3 branch. If there is any feedback feel free to add it to #63

PR #68, opened 2021-04-29
URI cleanup for JSON

stevespringett:
Ensure consistency for all URI/URLs to be defined as strings that must conform to the iri-reference format.

### DIFF
--- a/schema/bom-1.3-SNAPSHOT-strict.schema.json
+++ b/schema/bom-1.3-SNAPSHOT-strict.schema.json
@@ -172,19 +172,16 @@
       "properties": {
         "vendor": {
           "type": "string",
-          "format": "string",
           "title": "Tool Vendor",
           "description": "The date and time (timestamp) when the document was created."
         },
         "name": {
           "type": "string",
-          "format": "string",
           "title": "Tool Name",
           "description": "The date and time (timestamp) when the document was created."
         },
         "version": {
           "type": "string",
-          "format": "string",
           "title": "Tool Version",
           "description": "The date and time (timestamp) when the document was created."
         },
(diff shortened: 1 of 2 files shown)


coderpatros:
LGTM, do we have valid/invalid tests for those?

stevespringett:
> LGTM, do we have valid/invalid tests for those?

Good point. No, but let me add that.

stevespringett:
Valid test cases were committed with https://github.com/CycloneDX/specification/pull/71 which added support for hashes in references. The json test includes urls which passed validation. So in terms of backward compatibility, we should be good.

However, creating an invalid test seems to be problematic.  I haven't been able to get it to properly fail. According to https://json-schema.org/understanding-json-schema/reference/string.html:

> JSON Schema implementations are not required to implement this part of the specification, and many of them do not.

According to https://opis.io/json-schema/2.x/formats.html#iri-reference, the following should fail, but does not:

```json

{
  "bomFormat": "CycloneDX",
  "specVersion": "1.3",
  "serialNumber": "urn:uuid:3e671687-395b-41f5-a30f-a58921a69b79",
  "version": 1,
  "components": [
    {
      "type": "library",
      "publisher": "Acme Inc",
      "group": "org.example",
      "name": "mylibrary",
      "version": "1.0.0",
      "externalReferences": [
        {
          "type": "bom",
          "url": "\"\\\\\\\\WINDOWS\\\\filëßåré\""
        }
      ]
    }
  ]
}
```

If I change the JSON Schema to only accept uuid as the format, then validation fails like it should. I think most JSON schema validators have such a loose interpretation of iri-reference that everything basically passes. So I don't think it possible to add an invalid test case here.

coderpatros:
Working with JSON schema is so infuriating sometimes.

coderpatros (approved):

Issue #69, opened 2021-04-30
Add hash support to external references


stevespringett:
This was discussed many times but somehow never captured. External references need to optionally support integrity checking. While this may be irrelevant for URLs to websites or issue trackers, it could certainly be relevant to documentation (swagger doc), boms, and other remote resources.

PR #71, opened 2021-04-30
#69 - Added support for hashes on external references. Added unit tests

stevespringett:
Had to redo this PR as the prior one wasn't forked from v1.3, rather another branch causing issues.


coderpatros (approved):
LGTM

PR #74, opened 2021-05-04
Removed default empty string and unnecessary regex pattern

stevespringett:
Removed default empty string and unnecessary regex pattern that allowed all characters.

Java Core has already been updated with these changes. The checks performed in this PR will be using the modified versions which should validate they work.


coderpatros (approved):
LGTM

PR #75, opened 2021-05-04
Fix a few places where uri-reference has been applied at the array level instead of the array item level

coderpatros:


stevespringett (approved):
Good catch

PR #91, opened 2021-10-03
Added vulnerabilities as part of core spec

stevespringett:
This PR is based on work from #44 but without the limitations of the existing vulnerability extension. This implements #38.

Major changes from the vulnerability extension include:

* Each vulnerability now has its own bom-ref and can be referenced elsewhere in the BOM.
* Each vulnerability can now be applicable to 1 or more components or services. In the extension, there was a 1..1 mapping.
* A vulnerability can now reference zero or more aliases for the same vulnerability across different sources of vuln intel. For example, private vuln intel services typically have their own naming or numbering scheme. It is now possible to use these proprietary identifiers and still reference a CVE in the NVD, if the two vulnerabilities are in fact the same.
* Each rating now describes the source (e.g NVD, VulnDB, etc) as different sources may have different scores for the same vulnerability.
* The base, impact, and exploitable subscores were removed in favor of a single 'score' field. These were specific to CVSSv2 and CVSSv3 and are not applicable to OWASP. It is not clear if they will be applicable to CVSSv4. If these scores are wanted, use of the vector can recalculate the scores via automation, so no capabilities are lost.
* A justification field was added to provide additional context why something was rated the way it was.
* Vulnerability now has a details field which provides additional information over the 'description' field
* Recommendations was previously an array. This has changed to a single field and renamed 'remediation'
* Advisories were previously URLs. This has changed to being objects containing a title and URL.
* Created, published, and updated timestamp fields were added to the vulnerability
* Vulnerability now has a 'credits' field
* Vulnerability now has a 'tools' array which can describe the tools used to discover the vulnerability
* Vulnerability now has an 'analysis' object containing impact analysis of the vulnerability
* Vulnerability now includes the ability to utilize properties (introduced in v1.3) so that other data not supported by the schema can be added.


stevespringett:
Pinging @ddillard, @garethr, @wagoodman, @luhring, @planetlevel.

Your feedback on this (draft) PR is much appreciated. Once the details are flushed out, I'll update the PR to include XML and Protobuf schema support as well.

stevespringett:
Also pinging @brianf, @ahoog42, @llamahunter

Your feedback on this (draft) PR is much appreciated. Once the details are flushed out, I'll update the PR to include XML and Protobuf schema support as well.

stevespringett:
Also it should be noted that when you review this PR, look at it from two perspectives:
* Vulnerability information being in the same BOM as the components and other inventory
* Vulnerability information being externalized into a separate BOM, independent from inventory (e.g. a VEX BOM)

Both use cases will be possible which is exciting. 

For external use, since every component or service can have a reference to its own BOM, it would be possible to either:
* Have a single VEX BOM for the component the BOM describes (the root metadata component)
* Have multiple VEX BOMs for each and every component and service within the BOM.
* A combination of the two

Also note that I mentioned components or services. So also look at this PR from the perspective of:
* known vulnerabilities - identified by SCA and similar tools
* unknown vulnerabilities - identified from DAST and similar tools

In the future, we may want to add support for vulnerabilities from SAST and IAST/RASP tools, but that will require support for the different types of analyzers (e.g. semantic, control flow, data flow, etc) as well as the ability to support call graphs (including source and sinks). So although this PR will not address SAST today, please keep in mind that its not off the table for a future version of the spec.

VinodAnandan:
@knqyf263 @masahiro331 @raesene @samj1912 @mrutkows @nadgowdas @oliverchang @jlleitschuh @mlieberman85 @tjarrettveracode

schema/bom-1.4-strict-SNAPSHOT.schema.json
```diff
@@ -1074,6 +1082,278 @@
...
+    "impactAnalysisState": {
+      "type": "string",
+      "title": "Impact Analysis State",
+      "description": "Textual representation of the state of the impact analysis.",
+      "enum": [
+        "exploitable",
+        "in_triage",
+        "false_positive",
+        "not_affected",
+        "not_set"
```
stevespringett, line 1142: @coderpatros, @DarthHater Do we want a 'not_set' value here? Does this make sense or should we omit it?

brianf:
I looked through the spec but apologies if I missed it, is it possible to include mitigating information about the included vulns? eg "this thing has an old commons-collection cve but it only applies to unsafe deserialization and either this app doesn't have any serialization, or it's already whitelisted and so this can't be exploited here".

stevespringett:
>  is it possible to include mitigating information about the included vulns

Absolutely.  The spec proposes to include the following analysis states:
* exploitable
* false_positive
* not_applicable
* in_triage

In addition, both `impact` and `detail` fields are provided to provide context about why an analysis decision was made the way it was.

In the event a library was 'fixed'. This is already supported by the 'pedigree' capabilities implemented v1.2 of the spec where users can specify the commit and diffs along with the issues (defects, enhancements, security) that they've resolved in their custom version of a library.

ddillard:
What exactly does "exploitable" mean?  Does it mean that it's directly exploitable or both directly and indirectly exploitable?

An example might help: suppose I have an XML parser vuln, but I authenticate the sender of the XML and it's trusted so that vulnerability should not be directly exploitable.  However, suppose there's another vulnerability (perhaps one that I'm currently unaware of) and an attacker was able to bypass or otherwise subvert that authentication, and the attacker could then gain some additional privilege or knowledge by then exploiting the parser vuln.  Would the parser vuln be considered exploitable?

stevespringett:
@ddillard I think what you'll find is that organizations will say they're `not_affected` by this particular vulnerability and in the impact and detail fields they will likely try to communicate why. In this case because access control prohibits exploitability to only authenticated users.

The organization receiving the BOM will have an opportunity to either accept that analysis or push back on it. 

But to your point, it might be worth expanding on what `exploitable` means. 

For example:
* exploitable_confirmed - a PoC exists that demonstrates the exploitability of the vulnerability
* exploitable_unconfirmed - the vulnerability should theoretically be exploitable, but a PoC does not exist
* exploitable_directly - The vulnerable code is executed and is exploitable
* exploitable_indirectly - The vulnerable code is executed but mitigating controls prohibit exploitability

Thoughts?

nadgowdas:
Yes, not every user is equally affected by a common  single vulnerability. I believe some alignment with this vulnerability disclosure guide might help: https://github.com/ossf/oss-vulnerability-guide/blob/main/guide.md

schema/bom-1.4-strict-SNAPSHOT.schema.json
DarthHater: I'm more apt to say non existence of the `impactAnalysisState` is representative of that, but I suppose that someone could be explicit that it's not set, rather than just not create it?

schema/bom-1.4-strict-SNAPSHOT.schema.json
```diff
@@ -1074,6 +1082,278 @@
...
+        "OWASP",
+        "Other"
+      ]
+    },
+    "impactAnalysisState": {
+      "type": "string",
+      "title": "Impact Analysis State",
+      "description": "Textual representation of the state of the impact analysis.",
+      "enum": [
+        "exploitable",
```
stevespringett, line 1138: Should we handle direct and indirect exploitability?

Should we handle confirmed and unconfirmed exploitability?

schema/bom-1.4-strict-SNAPSHOT.schema.json
```diff
@@ -1074,6 +1082,278 @@
...
+        "in_triage",
+        "false_positive",
+        "not_affected",
```
stevespringett, line 1141: Should we handle a few choices on why a vulnerability is `not_applicable`?

Examples: code unreachable, vulnerable code not included (eg. with webpack), etc.

oliverchang:
Thanks for looping me in @VinodAnandan ! 

Would it be possible to align this more with the OSSF OSV vulnerability schema: https://ossf.github.io/osv-schema/ to enable easier interop? 

In particular, the ["affected"](https://ossf.github.io/osv-schema/#affected-fields) field lets us be a bit more descriptive about describing version/commit ranges (and this aligns well with the corresponding field in the [CVE 5.0 schema](https://github.com/CVEProject/cve-schema/blob/master/schema/v5.0/docs/versions.md) which adapted the OSV format).

stevespringett:
Hi @oliverchang 

Is the `affected.package` field required? I'm not sure how a BOM could properly support `affected.package` when it needs to support things like containers, an installable operating system in its entirety (e.g. Windows XP Embedded), firmware, a hardware device, or a file. But in the event a CycloneDX component has a purl, then affected.package could be determined for interop.

I think the version and range options are likely easily supportable and I think they would make good additions. I'll add them to the proposal.

It looks like the rest of the OSV spec is already covered by this proposal, so should be easy interop.

oliverchang:
> Hi @oliverchang
> 
> Is the `affected.package` field required? I'm not sure how a BOM could properly support `affected.package` when it needs to support things like containers, an installable operating system in its entirety (e.g. Windows XP Embedded), firmware, a hardware device, or a file. But in the event a CycloneDX component has a purl, then affected.package could be determined for interop.

No, it is not a required field. In cases where there is no PURL/package it can be left out. 

> 
> I think the version and range options are likely easily supportable and I think they would make good additions. I'll add them to the proposal.

Thank you! 

> 
> It looks like the rest of the OSV spec is already covered by this proposal, so should be easy interop.

Agreed :)

schema/bom-1.4-strict-SNAPSHOT.schema.json
```diff
@@ -1074,6 +1082,278 @@
...
+          "description": "Published advisories of the vulnerability if provided.",
+          "items": {
+            "$ref": "#/definitions/advisory"
+          }
+        },
+        "created": {
+          "type": "string",
+          "format": "date-time",
+          "title": "Created",
+          "description": "The date and time (timestamp) when the vulnerability was created."
```
DarthHater, line 1293: I think this could use maybe a tiny bit more information, like is this when it was created on the source of truth for it, or you know created by a slip up in code, etc...

I'm pretty sure I know it's the former, but for others consuming it would be cool to be a bit more clear.

schema/bom-1.4-strict-SNAPSHOT.schema.json
```diff
@@ -1074,6 +1082,278 @@
...
+        "updated": {
+          "type": "string",
+          "format": "date-time",
+          "title": "Updated",
+          "description": "The date and time (timestamp) when the vulnerability was last updated."
+        },
+        "credits": {
+          "type": "string",
+          "title": "Credits",
+          "description": "Individuals or organizations credited by with the discovery of the vulnerability."
```
DarthHater, line 1310: Should this be similar to how we treat authors or orgs in other places in the bom? I don't have a strong feeling but it's just something that I thought about.

stevespringett: Valid point. Clarified description in commit a04e72f06bb2b52b476f45cfce1c640f5aa3dbfb

schema/bom-1.4-strict-SNAPSHOT.schema.json
stevespringett: Valid point. Clarified description in commit a04e72f06bb2b52b476f45cfce1c640f5aa3dbfb

schema/bom-1.4-strict-SNAPSHOT.schema.json
stevespringett: Valid point. @coderpatros any strong opinion here? We could potentially make credits into an object that supports both OrganizationalEntity and OrganizationalPerson so that we cover:
* Organizations that take credit
* People that work for organizations that take credit
* People not representing a company that take credit

Do we want to have this be structured or unstructured data is the question.

schema/bom-1.4-strict-SNAPSHOT.schema.json
DarthHater: This sounds akin to consumption of the component, and slightly unrelated to a vulnerability existing in a component, or am I thinking about this wrong?

stevespringett: Thats correct. This is a VEX use case. The essence is that not every component is going to be exploitable. Some components may never be called, or the vulnerable method or function is never called, etc. This is really about trying to communicate, as a software supplier, the inherited vulnerabilities as well as whether or not consumers will be impacted by them.

It also allows IAST/RASP vendors, researchers, etc, to communicate with certainty, whether or not a vulnerability was exploitable or not.

masahiro331:
I'm masahiro331. May I join this discussion..?
I would like to ask this question in the context of SCA.

The recommendation component describes how to fix or work around a vulnerability, but it does not mechanically determine whether a patch will be provided in the first place.
So, how about defining the availability of patches?

stevespringett:
@masahiro331 Software patches can refer to different things depending on if you're a developer or operator of software. The term 'patch' is already used in CycloneDX to describe code modifications made to a component. When you refer to 'patches' are you referring to the availability of unaffected (safe) versions? If so, would you recommend this be a boolean field, or an array or range of alternative versions?

masahiro331:
@stevespringett  
Thank you for your reply. 

I referring to the unaffected version.
>  If so, would you recommend this be a boolean field or an array or range of alternative versions?

I think the modification state and the modified version array need to be defined separately.

I think enum is better for the modification state, not bool.
There are at least three states in the modification state.
- Fixed
- Will not fixed
- Not fixed (to be modified in the future)
The other states are possible, but they are considered to be a combination of the vulnerability impact.

At the very least, it would be better if the modified version is listed in the recommendation and can be parsed mechanically if possible. (This is a very difficult problem, so I don't think it is essential.)

The unaffected(safe) version is described in the recommendation, so it is not a problem.

However, it would be better to be able to parse them mechanically if possible. (This is a very difficult issue, so I don't think it's necessary at this time.

stevespringett:
@masahiro331 
> I referring to the availability of unaffected (safe) versions.

Good. I'll add support for that.

> I think the modification state and the modified versions need to be defined separately.

I'm a little confused on the word `modification` and `modified version`. CycloneDX already support modified components though it support of pedigree. See https://cyclonedx.org/use-cases/#pedigree. With pedigree, uses can communicate specifically what they've fixed (enhancement, defect, security issues) and how (commits, diffs). CycloneDX does not support a 'fixed' state without specifying more info.

I think 'wont fix' and 'not fixed' would possibly be good additions to the vulnerability analysis section, likely in relation to impact.

masahiro331:
@stevespringett 

> I think 'wont fix' and 'not fixed' would possibly be good additions to the vulnerability analysis section, likely in relation to impact.

Thank you!

I will check the spec again, as I misunderstood a bit about the perception of pedigree.

planetlevel:
> * known vulnerabilities - identified by SCA and similar tools
> * unknown vulnerabilities - identified from DAST and similar tools
> 
> In the future, we may want to add support for vulnerabilities from SAST and IAST/RASP tools, but that will require support for the different types of analyzers (e.g. semantic, control flow, data flow, etc) as well as the ability to support call graphs (including source and sinks). Or we could simply integrate more closely with SARIF. So although this PR will not address SAST today, please keep in mind that it's not off the table for a future version of the spec.

I'm confused.  Why couldn't we represent vulnerabilities from SAST and IAST tools in the BOM?  I wasn't thinking that the BOM would have all the details from a DAST/SAST/IAST tool... certainly not a fulll SARIF file... just the high level details.   We don't have any of that kind of detail for vulnerabilities in OSS components.  Can you explain the thinking here a bit more?

stevespringett:
@oliverchang I've added affected ranges based on node-semver syntax, and I've added an array of affected ranges. Supporting `git` directly doesn't quite fit with CycloneDX. It currently supports `vcs` for external references. For git specifically, the map of commit hashes makes sense, for centralized vcs, this doesn't make sense at all. It may also date the spec. Hopefully in 10+ years, something better than git will come along. So excluding git commit hashes, I think enough support is built into the spec now to make most use cases interoperable. Feedback appreciated.

schema/bom-1.4-strict-SNAPSHOT.schema.json
stevespringett: @planetlevel if a vulnerability is not exploitable, I'm thinking about adding a 'justification' in the event the analysis is set to 'not_affected'.

The choices I'm thinking of are:
* code_not_reachable - the code cannot be reached at runtime
* code_not_present - the code has been removed, tree-shaked (e.g. via webpack)
* function_not_enabled - the code requires a configurable option that is not set
* compiled_by_avoidance - a compiler flag was set/unset which prevented the introduction of the vulnerability
* dependency_not_met  - the vulnerability requires the presence of another component which is not present in the system.

Do you agree with these?
Are there others?

planetlevel: This is a really good list.  I think adding code_requires_environment, code_protected_at_runtime, and code_protected_at_perimeter would help.

My full (OCD-refactored) list would look like:
* code_not_present - the code has been removed, tree-shaked (e.g. via webpack)
* code_not_reachable - the vulnerable code is not invoked at runtime
* code_requires_configuration - exploitability requires a configurable option to be set/unset
* code_requires_dependency - exploitability requires a dependency that is not present
* code_requires_environment - exploitability requires a certain environment which is not present
* code_protected_by_compiler - exploitability requires a compiler flag to be set/unset
* code_protected_at_runtime - exploits are prevented at runtime
* code_protected_at_perimeter - attacks are blocked at network perimeter
* code_protected_by_mitigating_control - preventative measures have been implemented that reduce the likelihood and/or impact of the vulnerability

schema/bom-1.4-strict-SNAPSHOT.schema.json
```diff
@@ -1074,6 +1082,315 @@
...
+      "additionalItems": false,
+      "items": {
+        "$ref": "#/definitions/versionRange"
+      }
+    },
+    "versionRange": {
+      "type": "string",
+      "title": "Range",
+      "description": "A simplified version range specification derived from node-semver. A version range is a set of comparators which specify versions that satisfy the range. A comparator is composed of an operator and a version. The set of primitive operators is: < Less than, <= Less than or equal to, > Greater than, >= Greater than or equal to, and = Equal. If no operator is specified, then equality is assumed, so this operator is optional, but MAY be included. Comparators can be joined by whitespace to form a comparator set. Only a single comparator set is allowed per range specified in the BOM. No further node-semver syntax is supported. Semver 2.0 and higher is required.",
+      "examples": [ ">=1.2.7 <1.3.0", ">=1.2.7" ]
```
oliverchang, line 1393: In our experience, supporting ordering operators and more freedom in specifying ranges like this ends up complicating consumers of this data a lot and allows more room for ambiguity.

Would it be possible to change the representation here to something more like the one in CVE 5.0: https://github.com/CVEProject/cve-schema/blob/master/schema/v5.0/docs/versions.md or the [OSV schema](https://ossf.github.io/osv-schema/#affected-fields)? This was the result of countless discussions between the OSV team and the CVE WG.

schema/bom-1.4-strict-SNAPSHOT.schema.json
stevespringett: Added justifications and descriptions in b7e37c6f7125d6c5ee2cf03323144fe81608c438

schema/bom-1.4-strict-SNAPSHOT.schema.json
stevespringett: Resolved in commit f15b46296c3578e15d7c33148803068bb00bba19

schema/bom-1.4-strict-SNAPSHOT.schema.json
stevespringett: Description now states exploitable = direct or indirect exploitability.  No interest in supporting confirmed and unconfirmed at this time.

schema/bom-1.4-strict-SNAPSHOT.schema.json
```diff
@@ -1074,6 +1082,335 @@
...
+        "credits": {
+          "type": "string",
+          "title": "Credits",
+          "description": "Individuals or organizations credited by with the discovery of the vulnerability."
+        },
+        "tools": {
+          "type": "array",
+          "title": "Creation Tools",
+          "description": "The tool(s) used in the creation of the BOM.",
+          "items": {"$ref": "#/definitions/tool"}
```
luhring, line 1333: I know this isn't part of this PR currently — but what do you think about adding a "URL" field to the definition of `tool`? I'm thinking about cases where a document claims that _X tool_ found these results, and a user wants to try to reproduce the scan. (For Grype, the URL value would be something like `https://github.com/anchore/grype`.)

schema/bom-1.4-strict-SNAPSHOT.schema.json
```diff
@@ -1074,6 +1082,335 @@
...
+          "description": "The id of the vulnerability as defined by the risk scoring methodology. For example CVE-2019-15842 (of https://nvd.nist.gov/vuln/detail/CVE-2019-15842)"
+        },
+        "source": {
+          "$ref": "#/definitions/vulnerabilitySource",
+          "description": "The source that published the vulnerability."
+        },
+        "references": {
+          "type": "array",
+          "title": "References",
+          "description": "",
```
luhring, line 1246: I think a description here might help...

I had been wondering how to express that a vuln from one source (e.g. GHSA) refers to a vuln from another source (e.g. NVD), and I _think_ `references` is how one would do that — am I interpreting this correctly? Or would `bom-ref` be the mechanism to use?

luhring (commented):
Thanks for looping me in @stevespringett! Sorry for taking so long to get to this 🙇

schema/bom-1.4-strict-SNAPSHOT.schema.json
stevespringett: @luhring Good suggestion. I think this can easily be implemented by adding support for externalReferences to tools.. This would provide all kinds of different URLs to the tool including documentation, download, BOM location, etc.  I created #101 to track this enhancement.

schema/bom-1.4-strict-SNAPSHOT.schema.json
stevespringett: @coderpatros Any opinion on this one?

schema/bom-1.4-strict-SNAPSHOT.schema.json
```diff
@@ -1074,6 +1082,278 @@
...
+        "tools": {
+          "type": "array",
+          "title": "Creation Tools",
+          "description": "The tool(s) used in the creation of the BOM.",
+          "items": {"$ref": "#/definitions/tool"}
+        },
+        "analysis": {
+          "type": "object",
+          "title": "Impact Analysis",
+          "description": "An assessment of the impact and exploitability of the vulnerability.",
```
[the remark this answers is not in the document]
stevespringett: Its currently in the 'ratings' property. One or more ratings can be specified. So if you opted to use temportal and environmental criteria in CVSS, you'll get some of that. You can also use OWASP risk rating which is directly supported.

schema/bom-1.4-strict-SNAPSHOT.schema.json
oliverchang: @stevespringett please let me know what you think here, or if there are any concerns with adopting something closer to OSV. It would be really great if we could enable easier interop :)

schema/bom-1.4-strict-SNAPSHOT.schema.json
stevespringett: Yes @luhring. I've updated the description to state:

> Zero or more pointers to vulnerabilities that are the equivalent of the vulnerability specified. Often times, the same vulnerability may exist in multiple sources of vulnerability intelligence, but have a different identifiers. References provide a way to correlate vulnerabilities across multiple sources of vulnerability intelligence.

So your assumption was correct in its meaning.

schema/bom-1.4-strict-SNAPSHOT.schema.json
```diff
@@ -1074,6 +1082,335 @@
...
+        "detail": {
+          "type": "string",
+          "title": "Details",
+          "description": "If available, an in-depth description of the vulnerability as provided by the source organization. Details often include examples, proof-of-concepts, and other information useful in understanding root cause."
+        },
+        "recommendation": {
+          "type": "string",
+          "title": "Details",
+          "description": "Recommendations of how the vulnerability can be remediated or mitigated."
+        },
```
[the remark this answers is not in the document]
stevespringett: > 1. Should this field name get updated to remediation?

The spec supports `recommendation` which would typically include remediation or mitigation advise. 

CVE 5.0 schema supports 'workaround', which is kinda the same thing.

Do you see any issues with the current approach?

> 2. What's the motivation for collapsing the array into a string?

I was looking at the OSV schema and they have a single field. I also looked at CVE 5.0 schema and they support an array, but it appears to be used for languages - so a vulnerability supporting recommendations in English, French, and Mandarin for example.

Do you see value in supporting an array of recommendations? I'm indifferent about it.

schema/bom-1.4-strict-SNAPSHOT.schema.json
```diff
@@ -1074,6 +1082,278 @@
...
+    "severity": {
+      "type": "string",
+      "title": "Severity",
+      "description": "Textual representation of the severity of the vulnerability adopted by the risk analysis method. If an other risk analysis method is used other than whats defined in scoreSourceType, the user is expected to translate appropriately to match with an element value below.",
+      "enum": [
+        "critical",
+        "high",
+        "medium",
+        "low",
+        "none",
```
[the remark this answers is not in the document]
stevespringett: @luhring `none` is intended to be informational. Perhaps it should change to `info`? In some cases with previously unknown vulnerabilities against services, there's a need to report a informational finding that isn't technically a vulnerability. For example, the absence of Content Security Policy on a website to mitigate XSS, but no XSS was actually discovered. It's an informational warning.

I don't think it would ever be used for components however. I cannot see a use case for `info`, but perhaps there is one?

schema/bom-1.4-strict-SNAPSHOT.schema.json
stevespringett: Hi @oliverchang I have to admit, this is the hardest thing to design about this entire spec. So I haven't forgot about it - quite the opposite.  I'm not overly fond of the CVE 5.0 way of doing things - the spec is entirely too complex. I like the OSV way, but dislike the requirement on order. Introducing order into the spec when it wasn't previously a concern will be problematic.  

From my understanding reading the two specs and your comments, the request is to essentially make support for ranges 'statically typed' (for lack of a better phrase), so that the string doesn't have to be parsed. The precise meaning of the range will be conveyed throughout multiple fields with precise definitions.

Still working on it, but open to suggestions.

oliverchang: Thanks for taking a look at both CVE 5.0 and OSV @stevespringett !

Could you please clarify what you mean by "requirement on order"? OSV doesn't require the ranges to be specified in any particular order (though it's recommended for readability). The evaluation algorithm to determine if a version is included in the specified ranges requires an ordering (which would be the case with the current proposal as well because of the ordering operators).

Here's my pitch for adopting a more OSV like range definition :) 

From what I can tell, the way [OSV specifies ranges](https://ossf.github.io/osv-schema/#examples) would be *almost* equivalent to restricting the current range operators to only >= and <, as well as making things easier for more versioning schemes beyond SemVer. 

i.e. Instead of Supporting ">0.9.9 <1.3.0", ">= 1.0.0 <=1.2.9" etc with any combination of supported semver operators (>, <, >=, <=, =), OSV encodes versions as individual points on a "timeline" which indicates the first versions that contain the vulnerability, and the versions which fixes them. 

e.g. 

```json
"ranges": [ {
    "type": "SEMVER",
    "events": [
      { "introduced": "1.0.0" },
      { "fixed": "1.3.0" },  // order these are specified in doesn't matter here. 
    ]
} ]
```

Which is encoding ">=1.0.0 <1.3.0" (but only allowing these combination of operators and not requiring them to be grouped together). Encoding things this way also makes it clearer to consumers which versions contain patches. e.g. if a range is specified as "<=2.0.0-pre", what contains the patch? "2.0.0-pre.2", or "2.0.1"?

Or if there are multiple branches that contain the vulnerability,

```json
"ranges": [ {
    "type": "SEMVER",
    "events": [
      { "introduced": "1.0.0" },
      { "fixed": "1.3.0" }, 
      { "introduced": "2.0.0" }, 
      { "fixed": "2.5.0" }, 
    ]
} ]
```

This encodes >=1.0.0 <1.3.0 is vulnerable, and >=2.0.0 <2.5.0 is vulnerable.

This also generalises cleanly to GIT ranges and its non-linear history:

```json
"ranges": [ {
    "type": "GIT",
    "events": [
      { "introduced": "hash0" },  // common regressing commit
      { "fixed": "hash1" }, // fix in a branch
      { "fixed": "hash2" },  // fix in another branch
    ]
} ]
```

Finally, ecosystems that strictly follow SemVer is rare (AFAIK only Go, NPM do this in language ecosystems), and only supporting `node-semver` style ranges would limit its applicability. 

Not having explicit operators (which have different rules in different versioning schemes) in the OSV spec means we can also support these other ecosystems in exactly the same way:

```json
"ranges": [ {
    "type": "ECOSYSTEM",
    "events": [
      { "introduced": "1.0.0.0.0" },
      { "fixed": "2.0.0.0.post4" },
    ]
} ]
```

schema/bom-1.4-strict-SNAPSHOT.schema.json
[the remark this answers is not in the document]
luhring: >>Should this field name get updated to remediation?
>
>The spec supports recommendation which would typically include remediation or mitigation advise.

Just to clarify, my confusion is around this part of the PR's description:

>This has changed to a single field and renamed 'remediation'

I think logically the structure makes sense, this was just a minor naming curiosity.

---

>Do you see value in supporting an array of recommendations? I'm indifferent about it.

Good question. So far, only in contrived scenarios. I think string is good for the time being. 👍

schema/bom-1.4-strict-SNAPSHOT.schema.json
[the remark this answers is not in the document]
luhring: I don't have much of an opinion on this one. I just wanted to note that in some contexts, there might be a negligible (pun intended!) mis-translation from scanners' upstream data sources to the final report in CycloneDX, in the case where the upstream source didn't actually say "none". I think this is an unimportant distinction, for the most part, so there might be no need to change this.

schema/bom-1.4-strict-SNAPSHOT.schema.json
stevespringett: In the example:
```json
"events": [
      { "introduced": "1.0.0" },
      { "fixed": "3.0.0" }, 
      { "introduced": "2.0.0-rc1" }, 
      { "fixed": "2.5.0-rc5" }, 
    ]
```

if the ordering was not preserved and the json ended up being:

```json
"events": [
      { "introduced": "1.0.0" },
      { "fixed": "2.5.0-rc5" }, 
      { "introduced": "2.0.0-rc1" }, 
      { "fixed": "3.0.0" }, 
    ]
```    
It ends up being a bit of an odd overlap, and the consumer would need to perform version comparisons in business logic in order to make sense of this. 

It would be a bit more perplexing if the order came in as:

```json
"events": [
      { "introduced": "1.0.0" },
      { "introduced": "2.0.0-rc1" }, 
      { "fixed": "2.5.0-rc5" }, 
      { "fixed": "3.0.0" }, 
    ]
```    

With this data, I really don't know if 2.5.0-rc1 through rc4 are affected or not. So I'd like to avoid that kind of confusion in the spec.

I like the simplicity of OSV, but need to implement it in a way where business logic validation is not required, and do so in a way where we can support unaffected version ranges and unknown version ranges as well.

oliverchang: Avoiding ambiguity was a guiding principle of OSV :) The way these events are [evaluated](https://ossf.github.io/osv-schema/#evaluation) should have no room for ambiguity. 

Re 

```json
"events": [
      { "introduced": "1.0.0" },
      { "fixed": "2.0.0-rc5" }, 
      { "introduced": "2.0.0-rc1" }, 
      { "fixed": "3.0.0" }, 
    ]
```

The order these are specified does not matter for evaluation purposes, because the [evaluation algorithm](https://ossf.github.io/osv-schema/#evaluation) would sort this internally when processing this. i.e. The algorithm would first sort this as:

```json
"events": [
      { "introduced": "1.0.0" },
      { "introduced": "2.0.0-rc1" }, 
      { "fixed": "2.0.0-rc5" }, 
      { "fixed": "3.0.0" }, 
    ]
```

Which is your last example. Then the algorithm would say:

1.0.0 -> everything after this is affected 
until...
2.0.0-rc1 -> everything after this is affected (this ends up being redundant).
until...
2.0.0-rc5 -> everything after this is fixed
until...
3.0.0 -> everything after this is fixed (this is also redundant). 

So this ends up meaning the same thing as every version from 1.0.0 (inclusive) to 2.0.0-rc5 (exclusive) is affected. To answer your question specifically on if "2.0.0-rc1 through rc4 are affected", the answer would be yes.

Using the current node-semver encoding and preserving the order of your example, this would be:

```json
[
  ">=1.0.0 <3.0.0",
  ">=2.0.0-rc1 <2.0.0-rc5"
]
```

But this seems even more ambiguous to me. `2.0.0-rc5` is listed as an unaffected version for the second one, but it fits into the first range condition as well. Is it affected or not? Flattening this into a list of introduced/fixed versions avoids this. 

Specifying this as ">= introduced <fix" would also requires version comparison in business logic to understand these anyway right? The only way to truly avoid complicated version comparisons is to have explicit lists of versions rather than ranges (which is already a part of this current PR). However some ecosystems (such as Go, where every single commit maps to a ["pseudo-version"](https://golang.org/ref/mod#pseudo-versions)) don't make this feasible.

stevespringett: Thank you for the detailed explanation. I ended up porting the version range syntax from the CVE v5.0 schema to CycloneDX. My feeling is that there's less logic that would need to be implemented using that approach, even though I prefer the event driven approach of OSV. Anyway, I think for compatibility sake, CycloneDX is now perfectly aligned to CVE v5.0 and from my understanding there's a way to transpose to/from OSV. 

So I'm going to resolve this for now. Feel free to unresolve if necessary. Again, thanks so much for the guidance. Truly appreciated.

stevespringett:
@masahiro331 unaffected version ranges are now supported in the current draft.

schema/bom-1.4-strict-SNAPSHOT.schema.json
oliverchang: Thanks @stevespringett ! Aligning this to CVE v5.0 sounds great to me (and yes indeed it's compatible with OSV with very similar algorithms for evaluation).

schema/bom-1.4-strict-SNAPSHOT.schema.json
```diff
@@ -1074,6 +1082,278 @@
...
+        "unknown"
```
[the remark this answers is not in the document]
stevespringett: I've added an 'info' category in addition to the ones currently supported.

schema/bom-1.4-strict-SNAPSHOT.schema.json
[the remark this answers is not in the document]
stevespringett: I've added an 'info' category in addition to the ones supported.

stevespringett:
If the analysis state is `exploitable`, I'm thinking about including a `response`. The use case for this would be to document a vendors response to a vulnerability, which is in fact exploitable.

This is what I'm thinking. Thoughts?

```json
"response": {
    "type": "string",
    "title": "Response",
    "description": "A response to the vulnerability by the manufacturer, supplier, or project responsible for the affected component or service.",
    "enum": [
        "can_not_fix",
        "will_not_fix",
        "update",
        "rollback",
        "workaround_available"
    ]
}
```

stevespringett:
The majority of feedback and changes to this PR have already been worked out. I'll be working on getting the XML Schema and Protobuf schemas updated this week.

coderpatros:
In general I like this, but it seems like it has bit of a vulnerability centric view versus a vulnerability in context of an assembled piece of software view. An example to demonstrate.

I'm shipping ACME Widget Studio. CVE-EverythingIsOnFire, which affects component C, is published. Some segments of my dependency graph are as follows (X is ACME Widget Studio, * indicates a potentially exploitable component)...

1. X* -> A* -> B* -> C*
2. X -> D -> B* -> C*
3. X -> E -> C*
4. X* -> C*

The context of the CVE is quite different for each of those dependency graph segments.

I might not be groking this right. But I'm not sure how I would represent these different scenarios in the same BOM.

2 and 3 aren't exploitable, for different reasons, which I want to communicate.

1 and 4 are potentially exploitable, depending.

Do I include the same vulnerability multiple times for each?

Also, maybe, "affects" should be changed to "appliesTo"?

stevespringett:
@coderpatros This implementation currently takes a vulnerability per component(s) approach. In your example, component C is potentially exploitable in half of the usages of the component, and not exploitable in the other half.

Component C is the thing we're applying the vulnerability to, not the dependency relationship.

Will the consumer care if only half of the usages of component C are exploitable, or will they care if component C is exploitable?

> 2 and 3 aren't exploitable, for different reasons, which I want to communicate.
> 
> 1 and 4 are potentially exploitable, depending.
>
> Do I include the same vulnerability multiple times for each?

There currently isn't a way to communicate a call graph or node of a dependency graph. So currently, you'd have to include Component C multiple times and add the vulnerability multiple times with different scores and severity.

As stated in https://github.com/CycloneDX/specification/pull/91#issuecomment-932979748 I'd like to eventually incorporate SARIF support so that call graphs could be represented, which would provide much more detail beyond representing individual nodes of a dependency graph.

The question is, do we want to add support for dependency graphs now?

If so, are there proposals on how to do this that would not break backward compatibility?

stevespringett:
One approach may be to do something like:

```json
"affects": [
  {
    "ref": "pkg:maven/com.fasterxml.jackson.core/jackson-databind@2.9.9",
    "versions": [
      {
        "version": "2.9.0",
        "versionType": "semver",
        "lessThanOrEqual": "2.9.4",
        "status": "affected"
      }
    ],
    "constraints": [
      "appliesTo": [
        {
          "route": ["acme-x", "component-a", "component-b", "component-c"],
          "status": "affected"
        },
        {
          "route": ["acme-x", "component-d", "component-b", "component-c"],
          "status": "not_affected"
        },
        {
          "route": ["acme-x", "component-e", "component-c"],
          "status": "not_affected"
        },
        {
          "route": ["acme-x", "component-c"],
          "status": "affected"
        }
      ]
    ]
  }
]
```

If this is valuable information to represent, we could certainly do that. My only issue with this is that its too high level and focuses on the component, not the call graph. For example,

There may be multiple paths through `acme-x` -> `component-e` -> `component-c`, and only one leads to a vulnerability.
`"route": ["acme-x", "component-e", "component-c"],`

Representing the source and sink in a call stack would be highly valuable in that case, but would require us to integrate with SARIF.

stevespringett:
@coderpatros, @DarthHater, any feedback on https://github.com/CycloneDX/specification/pull/91#issuecomment-976127606?

schema/bom-1.4-strict-SNAPSHOT.schema.json
stevespringett: CycloneDX Core meeting - Preference on structured data rather than a string field.

stevespringett:
Per CycloneDX Core meeting - make 'response' as shown [here](https://github.com/CycloneDX/specification/pull/91#issuecomment-976127606) an array.

schema/bom-1.4-strict-SNAPSHOT.schema.json
stevespringett: This was changed to structured data in b42bbca0ba3196c5c4ae204155b6352f85970cc5

stevespringett:
Migrate version range syntax to Package URL version range syntax (vers)
https://github.com/package-url/purl-spec/pull/139

Currently, it's based on CVE 5.0 schema, which has a number of problems including an undefined way to specify version range type.

schema/bom-1.4-SNAPSHOT.proto
```diff
@@ -450,3 +452,179 @@ message Evidence {
...
+
+enum Severity {
+  SEVERITY_CRITICAL = 0;
+  SEVERITY_HIGH = 1;
+  SEVERITY_MEDIUM = 2;
+  SEVERITY_LOW = 3;
+  SEVERITY_INFO = 4;
+  SEVERITY_NONE = 5;
+  SEVERITY_UNKNOWN = 6;
+}
```
coderpatros, line 523: With how protobuf handles default values, like 0, it can catch out people implementing serialization and deserialization.

I would make the default 0 value SEVERITY_UNKNOWN for the severity enum.

schema/bom-1.4-SNAPSHOT.proto
```diff
@@ -450,3 +452,179 @@ message Evidence {
...
+  SCORE_METHOD_CVSSV2 = 0;
+  // Common Vulnerability Scoring System v3 - https://www.first.org/cvss/v3-0/
+  SCORE_METHOD_CVSSV3 = 1;
+  // Common Vulnerability Scoring System v3.1 - https://www.first.org/cvss/v3-1/
+  SCORE_METHOD_CVSSV31 = 2;
+  // OWASP Risk Rating Methodology - https://owasp.org/www-community/OWASP_Risk_Rating_Methodology
+  SCORE_METHOD_OWASP = 3;
+  // Other scoring method
+  SCORE_METHOD_OTHER = 4;
+}
```
coderpatros, line 536: If might be useful, similar reasons to above, to include SCORE_METHOD_NULL as the default 0 value.

schema/bom-1.4-SNAPSHOT.proto
```diff
@@ -450,3 +452,179 @@ message Evidence {
...
+  IMPACT_ANALYSIS_STATE_RESOLVED_WITH_PEDIGREE = 1;
+  // The vulnerability may be directly or indirectly exploitable.
+  IMPACT_ANALYSIS_STATE_EXPLOITABLE = 2;
+  // The vulnerability is being investigated.
+  IMPACT_ANALYSIS_STATE_IN_TRIAGE = 3;
+  // The vulnerability is not specific to the component or service and was falsely identified or associated.
+  IMPACT_ANALYSIS_STATE_FALSE_POSITIVE = 4;
+  // The component or service is not affected by the vulnerability. Justification should be specified for all not_affected cases.
+  IMPACT_ANALYSIS_STATE_NOT_AFFECTED = 5;
+}
```
coderpatros, line 576: As above, IMPACT_ANALYSIS_STATE_NULL as default 0 value.

schema/bom-1.4-SNAPSHOT.proto
```diff
@@ -450,3 +452,179 @@ message Evidence {
...
+  IMPACT_ANALYSIS_JUSTIFICATION_PROTECTED_BY_MITIGATING_CONTROL = 8;
+}
+
+enum VulnerabilityResponse {
+  VULNERABILITY_RESPONSE_CAN_NOT_FIX = 0;
+  VULNERABILITY_RESPONSE_WILL_NOT_FIX = 1;
+  VULNERABILITY_RESPONSE_UPDATE = 2;
+  VULNERABILITY_RESPONSE_ROLLBACK = 3;
+  VULNERABILITY_RESPONSE_WORKAROUND_AVAILABLE = 4;
+}
```
coderpatros, line 605: More default "null" values suggested

schema/bom-1.4-SNAPSHOT.proto
```diff
@@ -450,3 +452,179 @@ message Evidence {
...
+  // The vulnerability status for the version or range of versions.
+  optional VulnerabilityAffectedStatus status = 3;
+}
+
+enum VulnerabilityAffectedStatus {
+  // The vulnerability status of a given version or range of versions of a product. The statuses 'affected' and 'unaffected' indicate that the version is affected or unaffected by the vulnerability. The status 'unknown' indicates that it is unknown or unspecified whether the given version is affected. There can be many reasons for an 'unknown' status, including that an investigation has not been undertaken or that a vendor has not disclosed the status.
+  VULNERABILITY_AFFECTED_STATUS_AFFECTED = 0;
+  VULNERABILITY_AFFECTED_STATUS_NOT_AFFECTED = 1;
+  VULNERABILITY_AFFECTED_STATUS_UNKNOWN = 2;
+}
```
coderpatros, line 630: Shift UNKNOWN to default 0 value

schema/bom-1.4-SNAPSHOT.proto
stevespringett: Good catch. Corrected in 937b8fd6d0f7856f4f2d92b6e3b99102386125d8

schema/bom-1.4-SNAPSHOT.proto
stevespringett: Corrected in 7a7d7ada967bf138be476c68c1072c2f0696bc78

schema/bom-1.4-SNAPSHOT.proto
stevespringett: Changed in 6201ede37ac9c730acfe414073fa281ab900ad98

schema/bom-1.4-SNAPSHOT.proto
stevespringett: Changed in c85bdd66a3fe7413bdc31c73dd9347acdb01860d

schema/bom-1.4-SNAPSHOT.proto
stevespringett: Changed in 0e1df61cef885ce82273b5afa0295358e1ca9c5a

coderpatros (approved):

Issue #89, opened 2021-09-30
Harden JSON Schema


stevespringett:
The JSON schema, by default, is relaxed and allows undefined properties to appear anywhere in the document. An optional strict schema is provided that disallows this behavior.

This proposal would:
* eliminate the relaxed schema
* JSON schema would not allow undefined properties to appear (strict)

Impacts:
* JSON serialization format would not provide a playground for extensions. This may not be overly impactful as support for Properties were introduced in v1.3 which eliminates most of the need for extensions. Additionally, JSON extensions were problematic anyway as JSON Schema is not extensible in the same way XML is.
* Enveloped signing will need to be specified as part of the JSON spec (not the XML spec???). This is due to JSON not supporting namespaces. Therefore we will need to define the parts of the SBOM that are the most likely to be signed.  This will likely be scoped to:
  * bom
  * component
  * service

coderpatros:
I think we should add it to the xml spec too. Be good to have a hardened xml spec too.

coderpatros:
I think we should add it to the xml spec too. Be good to have a hardened xml spec too.

stevespringett:
I've reached out to NSA to see if they've evaluated CycloneDX. I know it was on their todo. Not sure if they've gotten around to it yet.

stevespringett:
I heard back from Anders Rundgren, the author of JSF from webpki.org and a JSON schema does not exist for JSF at the moment. However, his recommendations would be to create a subset of JSF for our use cases. As long as we require [RFC 8785](https://datatracker.ietf.org/doc/html/rfc8785), which he recently got through, we should be good.

PR #93, opened 2021-10-18
Implemented JSF in the core spec

stevespringett:
This implements a prerequisite in #89


coderpatros (approved):

coderpatros:
LGTM

I can't remember, but I think you were creating the JSF schema from scratch? I'm wondering if we should [at]cyberphone to see if he could review it too?

stevespringett:
HI @cyberphone. As stated on the prior email thread, we're implementing JSF 0.82 directly into CycloneDX. Would you mind taking a look?

The schema supports the entirety of the JSF spec excluding `extensions`. The way extensions are currently implemented in JSF requires JSON properties to be dynamic and thus the JSON Schema `additionalProperties` value must be set to true, which prohibits the schema from being hardened. So we opted not to support JSF extensions so that our schema could remain hardened. 

Let us know if we captured everything correctly or if there's something we need to change.

Also, do you see any major compatibility issues with excluding extension support?

cyberphone:
I'm impressed!  BTW, I didn't know that JSON schema was that powerful.

If `extensions` is important or not is very application dependent.  Its main purpose is holding additional signatory meta data if there are multiple signers like in https://cyberphone.github.io/doc/security/jsf.html#p256_es256_r2048_rs256_mult-exts-kid.json

I will most certainly reference this work in other contexts.

stevespringett:
@DarthHater can I get approval on this please?

PR #120, opened 2022-01-03
Add service release notes to v1.4 proto file

coderpatros:
This was missing from the proto file, but is present in JSON and XML schemas.


stevespringett:
Nice catch.

PR #121, opened 2022-01-08
v1.4 General Availability

stevespringett:


coderpatros (approved):
+1 from me.

As mentioned on slack, I've done a feature complete implementation of v1.4 and done some performance regression tests. This is in good shape for release IMHO.

PR #144, opened 2022-05-01
fix CWEs example

### LINKED ISSUE #143: v1.4 vulnerabilities[].cwes spec/example mismatch
The spec says `vulnerabilities[].cwes` should be an array of integers, but the example shows a string.

image

Compare for example with `version` which correctly shows an integer example.

image

kabo:
Closes #143


stevespringett:
Thanks for the correction. Much appreciated.

PR #313, opened 2023-10-11
CBOM: merges relatedCryptoMaterial and key asset types

bhess:
This addresses the use case described in https://github.com/IBM/CBOM/issues/31:
- Expiry and life cycle of any relatedCryptoMaterial (e.g. keys, tokens, password) should be expressible.

This is done by merging the "key" asset type and "relatedCryptoMaterial" to "relatedCryptoMaterial", which contains the needed properties.


bhess:
The property that caused the schema validation to fail is updated.

PR #339, opened 2023-11-16
CBOM: adds 'parameterSetIdentifier' property, replacing 'variant'

bhess:
The new property 'parameterSetIdentifier' replaces 'variant' and contains information about the parameter set identifying an algorithm. This can be, for example, the key length (in AES), the digest length (in SHA2), or the hash algorithm used internally (in SLH-DSA / FIPS205). The "description" field contains some examples.

This PR is motivated by https://github.com/IBM/CBOM/issues/37 and intends to address its use case.

Tagging @stevespringett, @n1ckl0sk0rtge, @mrutkows, @GeroDittmann

### DIFF
--- a/schema/bom-1.6.schema.json
+++ b/schema/bom-1.6.schema.json
@@ -3879,15 +3879,10 @@
                 "unknown"
               ]
             },
-            "variant": {
+            "parameterSetIdentifier": {
               "type": "string",
-              "title": "variant",
-              "description": "The variant for the crypto algorithm, algorithm components should be delimited with '-'",
-              "examples": [
-                "aes128-cbc-pkcs7",
-                "rsa2048-cbc-oaep-sha256-mgf1",
-                "kyber1024"
-              ]
+              "title": "parameter set identifier",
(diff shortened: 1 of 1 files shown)


GeroDittmann:
Thanks for the tag, @bhess . Could you elaborate a little how this would be used?

In cases where a user or policy might care for a key length, would they typically find it in the `title` field? Without parsing a string?

Does the `title` field identify a parameter set, as `parameterSetIdentifier` suggests? Are these identifiers formally defined somewhere?

If I understand the examples in the description correctly, the interpretation of the `title` field depends on the algorithm it belongs to?

bhess:
Hi @GeroDittmann, 

> In cases where a user or policy might care for a key length, would they typically find it in the `title` field? Without parsing a string?

The 'title' field is just for the schema definition, for informational purposes. The user would just set the field 'parameterSetIdentifier'. In the AES case this will be the key length, see the "description" part for some examples.



> If I understand the examples in the description correctly, the interpretation of the `title` field depends on the algorithm it belongs to?

The 'parameterSetIdentifier' field depends on the algorithm it belongs to.

We don't formally define them (e.g. in terms of an exhaustive enum). There would need to be an authoritative source for such a list, which I think is outside the scope of the CBOM.

PR #353, opened 2023-12-07
Support for hybrids/combiners: add 'combiner' as primitive

bhess:
Adds 'combiner' as enum entry in 'primitive'.

Addresses the use case when combinations of (e.g.) classical crypto like RSA is used together with QSC like Dilithium. An entry of this primitive allows to express the combiner used. Adding dependencies to RSA/Dilithium then allows to express the algorithms used in the combiner. Note: "combiners" are also known as "hybrids", but this term can be ambiguous so I prefer the term combiner.

### DIFF
--- a/schema/bom-1.6.schema.json
+++ b/schema/bom-1.6.schema.json
@@ -3880,7 +3880,7 @@
             "primitive": {
               "type": "string",
               "title": "primitive",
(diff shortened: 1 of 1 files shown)


bhess:
Signed off the commit now to make CI pass.

PR #395, opened 2024-03-07
Propose new environmental consideration information for ML models

mrutkows:
The fact that datasets used to train AI models are increasingly large and take an enormous amount of energy (and indirectly produce large CO2 emissions) to develop, train and run has come to the forefront.  This PR contains proposed additions to the "modelCard" type to account for these considerations when selecting/utilizing a model.

- Adds `ModelCardConsiderations.environmentalConsiderations` 
  this fixes https://github.com/CycloneDX/specification/issues/396
- Adds `OrganizationalEntity.address`

----

TODO

- [x] modify JSON schema
- [x] modify XML schema
- [x] modify protobuf schema


mrutkows:
**NOTE**: we need to add "externalReferences' and "properties" where applicable

jkowalleck:
@mrutkows is this PR related to an existing issue? 
If not, please create an issue that describes the reasoning, use cases, edge cases, out-of-scopes, and such.

jkowalleck:
this issue claims to solve #396

well, since the #396 does in no way describe a problem yet, I cannot see how this PR intends to solve it. 
@mrutkows , please take the 10 minute to describe the actual problem, before I take the two hours to actually review the "solution"

stevespringett:
@mrutkows I think we have consensus to use the current model card approach to adding support for environmental concerns and then in v1.7, we can expand that support to every component and service.

What is the status of this PR? We likely have one week to flush this out, otherwise it will need to be postponed to v1.7.

jkowalleck:
I still have questions and concerns. I vote for postponing and continued discussions in #396 
If the data is needed earlier, we still could create CDX properties for this - and ses how it is adopted.

jkowalleck:
@mrutkows , Vinod and I had a discussion and will work on having this moved forward, having this part of 1.6

jkowalleck:
a followp will be https://github.com/CycloneDX/specification/issues/406

schema/bom-1.6.proto
```diff
@@ -1150,9 +1152,136 @@ message ModelCard {
...
+      optional CO2MeasureType co2CostOffset = 5;
+    }
   }
 }
 
+// An activity that is part of a machine learning model development or operational lifecycle.
+enum ActivityType {
+  // a lifecycle activity type whose description does not match currently defined values.
+  // buf:lint:ignore ENUM_ZERO_VALUE_SUFFIX -- `other` is our fallback
+  ACTIVITY_TYPE_OTHER = 0;
```
jkowalleck, line 1178: pleasse do not use `ACTIVITY_TYPE_OTHER ` as the default value.
better create `ACTIVITY_TYPE_UNSPECIFIED` for that purpose.\


the things where you have seen the `  // buf:lint:ignore ENUM_ZERO_VALUE_SUFFIX -- `other` is our fallback` were added to suppress the messache in acknowledgement of a mistake in the past.

schema/bom-1.6.proto
```diff
@@ -1150,9 +1152,136 @@ message ModelCard {
...
+  // energy source: wind
+  ENERGY_SOURCE_TYPE_WIND = 6;
+  // energy source: solar
+  ENERGY_SOURCE_TYPE_SOLAR = 7;
+  // energy source: geothermal
+  ENERGY_SOURCE_TYPE_GEOTHERMAL = 8;
+  // energy source: hydropower
+  ENERGY_SOURCE_TYPE_HYDROPOWER = 9;
+  // energy source: other; not yet defined
+  ENERGY_SOURCE_TYPE_OTHER = 10;
```
jkowalleck, line 1236: please move the `ENERGY_SOURCE_TYPE_OTHER` to value `1`

schema/bom-1.6.proto
```diff
@@ -1150,9 +1152,136 @@ message ModelCard {
...
+  // Quantity of carbon dioxide (CO2).
+  float value = 1;
+  // Unit of carbon dioxide (CO2).
+  CO2MeasureUnitType unit = 2;
+}
+
+enum CO2MeasureUnitType {
+  // Tonnes (t) of carbon dioxide (CO2) equivalent (eq).
+  // buf:lint:ignore ENUM_ZERO_VALUE_SUFFIX -- `tCO2eq` is our fallback
+  CO2_MEASURE_UNIT_TYPE_TONNES_CO2_EQUIVALENT = 0;
```
jkowalleck, line 1264: value `0` means - unset - which is equal to a "default" value.

therefore, you MUST NOT use it as such here.

jkowalleck (changes requested):

schema/bom-1.6.proto
mrutkows: Hey Jan, so a receiver of this message would have to map "unspecified" to "other"?  or should we actually define an enum value unspecified for all new enum. types (i.e., an actual value "unspecified" in the JSON and XML?

jkowalleck: "other" is a well-known case, according to JSON and XML.

But the value `0`, in Brotobuf v3(`proto3` schema), is the default value which is assigned if no value is given (for optional messages for example) and it is also the value that is assigned if a non-wellnown-value is received.
the case `...UNSPECIFIED = 0` is a protobuf3 thing only - no need to carry this design over to XML/JSON.

jkowalleck: read https://protobuf.dev/programming-guides/dos-donts/#unspecified-enum

schema/bom-1.6.proto
mrutkows: It would be desirable for the default to be "other".

jkowalleck: I do not think so
https://protobuf.dev/programming-guides/dos-donts/#unspecified-enum
> It may be tempting to declare this default as a semantically meaningful value but as a general rule, do not, to aid in the evolution of your protocol as new enum values are added over time.

schema/bom-1.6.proto
jkowalleck: each time you think of a default here, be aware that you need to keep it constisten with the JSON and XML.
I dont see a default value for the enum there, either.

schema/bom-1.6.xsd
```diff
@@ -3237,6 +3244,333 @@ limitations under the License.
...
+                <xs:annotation>
+                    <xs:documentation>
+                        Describes energy consumption information incurred for one or more component lifecycle activities.
+                    </xs:documentation>
+                </xs:annotation>
+            </xs:element>
+        </xs:sequence>
+    </xs:complexType>
+
+    <xs:complexType name="energyConsumptionType">
```
jkowalleck, line 3264: the values are very much ML-specific.

please prefix the element name with something appropriate. like `mlEnergyConsumptionType`

schema/bom-1.6.schema.json
```diff
@@ -3470,6 +3475,215 @@
...
+          "title": "Energy Consumptions",
+          "description": "Describes energy consumption information incurred for one or more component lifecycle activities.",
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/energyConsumption"
+          }
+        }
+      }
+    },
+    "energyConsumption": {
```
jkowalleck, line 3494: the values of `activity` are very ML-specific.
therefore, i think a name prefix would be suitable. like `mlEnergyConsumption`

schema/bom-1.6.proto
```diff
@@ -1150,9 +1152,136 @@ message ModelCard {
       // With respect to the benefits and harms outlined, please describe any mitigation strategy implemented.
       optional string mitigationStrategy = 4;
     }
+    message EnvironmentalConsiderations {
+      // Describes energy consumption information incurred for one or more component lifecycle activities.
+      repeated EnergyConsumption energyConsumptions = 1;
```
jkowalleck, line 1157: all the new messages and enums appear to be much centric to the existing message `ModelCardConsiderations`. especially `ActivityType` is ML-specific.
maybe nest those declarations in the message, instead of having them free floating somewhere.  like so: 
```
message Foo {
  message Bar { ... }
  enum Bazz { ... }
  message FooBar {
    repeated Foo foo = 1
  }

  optional Bar bar = 1
  FooBar foobar = 2
}
```

schema/bom-1.6.proto
jkowalleck: all done and fixed :+1:

schema/bom-1.6.proto
mrutkows: @jkowalleck I was trying to consider which structures would be reusable independent of model card considerations (and perhaps, if time allowed, consider adding a means to declare them in the parent model card so that we have a means to allow these concepts to be used by others in v1.7 as Prabhu indicated interest in).

mrutkows:
Please note that I added a new folder to hold "examples" under `docgen` (with one example so far) as it seems these should become part of the automated doc. generation.

schema/bom-1.6.schema.json
mrutkows: I agree that these values are ml-specific today, but to build on my thoughts above, other CI-CD related activities could be added in the future and then backing off the "ml" name would be more challenging.

mrutkows:
For those following, here is a snapshot of the (JSON) model visualization for just the subtree with the environmental consideration types:

Screenshot 2024-03-26 at 12 04 57 PM

Please note that I did not expand the postal address type above.

schema/bom-1.6.proto
jkowalleck: I disagree. 

i will drop a PR to show what is possible.

schema/bom-1.6.proto
jkowalleck: done

schema/bom-1.6.proto
jkowalleck: done

jkowalleck (changes requested):
test resources still missing.

the one example in `docgen` is never applied to any test, so i do not count this one.

schema/bom-1.6.xsd
stevespringett: @jkowalleck we should not use `mlEnergyConsumptionType` if this is to be used elsewhere. I agree with the current name of `energyConsumptionType`.

schema/bom-1.6.schema.json
stevespringett: This is not a blocking issues IMO.

mrutkows:
> test resources still missing.
> 
> the one example in `docgen` is never applied to any test, so i do not count this one.

Where can I find an example of a test reference?  My belief was that the CI automation linted and checked for format?

jkowalleck:
@mrutkows here is the neeed fix for the examples: https://github.com/mrutkows/specification/pull/3 

as soon as this is merged, we should be golden.

mrutkows:
> @mrutkows here is the neeed fix for the examples: [mrutkows#3](https://github.com/mrutkows/specification/pull/3)
> 
> as soon as this is merged, we should be golden.

Thanks Jan!!!

BTW, there is still something wrong in the XSD in that I should have an `energyConsumptionsType` (plural) as well as the current `energyConsumptionType` to account for the corresponding anonymous type in the JSON schema... I will commit that change shortly.

docgen/examples/md/mlbom-1.6-env-considerations-1.json
jkowalleck, line 1: the content of the file is invalid to schema, if i am not mistaken.
anyway, the file shall be removed.

jkowalleck (approved):
looks good overall.

schema/bom-1.6.schema.json
```diff
@@ -3470,6 +3480,240 @@
...
+        "energyProviders",
+        "activityEnergyCost"
+      ],
+      "additionalProperties": false,
+      "properties": {
+        "activity": {
+          "type": "string",
+          "title": "Activity",
+          "description": "The type of activity that is part of a machine learning model development or operational lifecycle.",
+          "enum": [ "design", "data-collection", "data-preparation", "training", "fine-tuning", "validation", "deployment", "inference" ],
```
stevespringett, line 3521: @mrutkows the enum is missing `other`. The `meta:enum` describes `other`.

stevespringett (commented):
This looks good @mrutkows . Thanks for all the hard work on this. If you can sort out the `other` discrepancy, we can get this approved.

schema/bom-1.6.schema.json
jkowalleck: @stevespringett this was not fixed before merging.

schema/bom-1.6.proto
jkowalleck: done

schema/bom-1.6.schema.json
jkowalleck: done

schema/bom-1.6.schema.json
jkowalleck: nevermind, was fixed via 197d1065426b666178683ff194700a40b28581e1

PR #412, opened 2024-03-24
Added support for concluded value. Updated test cases.

### LINKED ISSUE #411: Add concluded value to identity evidence
Identity evidence has many methods, each with a technique, confidence score, and associated value of the evidence. What is missing, is the concluded value from all of the methods.

For example:
```json
"evidence": {
  "identity": [
    {
      "field": "cpe",
      "confidence": 1.0,
      "concludedValue": "cpe:2.3:a:example:acme-application:1.0.0:*:*:*:*:*:*:*",
      "methods": [
        {
          "technique": "filename",
          "confidence": 0.1,
          "value": "acme-application-1.0.0.exe"
        },
        {
          "technique": "hash-comparison",
          "confidence": 0.8,
          "value": "7c547a9d67cc7bc315c93b6e2ff8e4b6b41ae5be454ac249655ecb5ca2a85abf"
        }
      ]
    }
  ]
}
```

The example above concludes that the combination of the two methods results in a specific CPE. The proposal is to add `concludedValue` so that this can be expressed.

stevespringett:
Closes #411 

- [x] modify JSON schema
- [x] modify XML schema
- [x] modify protobuf schema

### DIFF
--- a/schema/bom-1.6.schema.json
+++ b/schema/bom-1.6.schema.json
@@ -4441,6 +4441,11 @@
           "title": "Confidence",
           "description": "The overall confidence of the evidence from 0 - 1, where 1 is 100% confidence."
         },
+        "concludedValue": {
+          "type": "string",
+          "title": "Concluded Value",
+          "description": "The value of the field (cpe, purl, etc) that has been concluded based on the aggregate of all methods (if available)."
+        },
         "methods": {
           "type": "array",
           "title": "Methods",
--- a/schema/bom-1.6.proto
+++ b/schema/bom-1.6.proto
@@ -729,6 +729,8 @@ message EvidenceIdentity {
   repeated EvidenceMethods methods = 3;
   // The object in the BOM identified by its bom-ref. This is often a component or service, but may be any object type supporting bom-refs. Tools used for analysis should already be defined in the BOM, either in the metadata/tools, components, or formulation.
   repeated string tools = 4;
+  // The value of the field (cpe, purl, etc) that has been concluded based on the aggregate of all methods (if available).
+  optional string concludedValue = 5;
 }
 
 message EvidenceMethods {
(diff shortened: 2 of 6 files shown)


jkowalleck (approved):

schema/bom-1.6.schema.json
```diff
@@ -4441,6 +4441,11 @@
           "title": "Confidence",
           "description": "The overall confidence of the evidence from 0 - 1, where 1 is 100% confidence."
         },
+        "concludedValue": {
```
prabhu, line 4444: Isn't the concluded value already set for the given attribute? We are now creating an opportunity for an edge case where the concluded value could be different from the actual value for the given attribute.

stevespringett: > Isn't the concluded value already set for the given attribute?

No, this is an omission. 

> We are now creating an opportunity for an edge case where the concluded value could be different from the actual value for the given attribute.

Yes and no.  Yes, it it possible to have a purl for example in the `evidence.identity.method.value` field and it can be different than the `evidence.identity.concludedValue`. This is by design, as each method has a confidence. So you could have a high and low confidence purl, and your concludedValue might be the high confidence value.

If the `evidence.identity.method.value` is a hash for example, then the `concludedValue` might be a purl based on a dictionary lookup in a table for that hash value.  So in this case, the value and the concludedValue would be talking about two different things.

In either case, the `concludedValue` substantiates the identity in `component`. These two values should be the same, yes. But the `concludedValue` may be low confidence, therefore it instructs the end user that the asserted identity in `component` may be incorrect.

One common use case for doing it this way is for decoupling the SBOM and the evidence.  With the use of an external reference of type `evidence` in the SBOM, the SBOM could point to a BOM containing just the evidence for a given component.

PR #630, opened 2025-05-01
Added citation support and test cases.

stevespringett:


schema/bom-1.7.schema.json
```diff
@@ -5725,6 +5734,60 @@
...
+          "description": "An optional description or comment about the context or quality of the data attribution."
+        },
+        "signature": {
+          "$ref": "#/definitions/signature",
+          "title": "Signature",
+          "description": "An optional digital signature verifying the authenticity or integrity of the attribution."
+        }
+      },
+      "required": ["pointers", "timestamp"],
+      "anyOf": [
```
jkowalleck, line 6178: :x:  
what is this `anyOf` supposed to mean?
if i understand it correct, it means, atleast one of the following is required. ...

I missed a respective part in the XML and in the ProtoBuf - I would expect to see at least a text expressing this requirement.

i suggest 
- adding a textual annotation the ProtoBuf that expresses that rule
- adding a textual annotation the XML that expresses that rule. or add a `choice` if technically possible. 
- providing proper examples/test-data for valid and invalid showcases

jkowalleck (changes requested):

jkowalleck:
RFC notice sent on May 1, 2025

* https://groups.io/g/CycloneDX/message/309
* https://cyclonedx.slack.com/archives/CVA0G10FN/p1746132948459179

Public RFC period ended May 29, 2025

jkowalleck:
@stevespringett i see lacks in the implementation. I'd reject the current version for its unclear implementation.

schema/bom-1.7.schema.json
stevespringett: In https://github.com/CycloneDX/specification/pull/630/commits/be626a971dac652728d3c08a272c1c0419e67a6b the XSD and Proto were updated to include the requested documentation. Additionally, an invalid test case was created.

jkowalleck:
my remark was clarified. since there is nothing left in the spec, this is ready for TC54 vote.

i will fix the current merge conflicts, and i will add additional valid/invalid examples according to spec, and might adjust the schemas to detect the invalid cases if possible.

jkowalleck:
after reading this spec again, i really do not like it. :-1: 

the idea of `pointer` is a horror for most implementations that use (unsorted) sets for data storage. the order of elements never really mattered, but now it does.
this spec is much to much dependent if schema implementations (XML/JSON/PB) and programming-language implementations.

schema/bom-1.7.xsd
```diff
@@ -8856,6 +8861,105 @@ limitations under the License.
...
+                    <xs:documentation>
+                        The `bom-ref` of an object, such as a component, service, organisational entity, or person that supplied the cited information.
+                        At least one of the "attributedTo" or "process" elements must be present.
+                    </xs:documentation>
+                </xs:annotation>
+            </xs:element>
+            <xs:element name="process" type="bom:refLinkType" minOccurs="0" maxOccurs="1">
+                <xs:annotation>
+                    <xs:documentation>
+                        The `bom-ref` to a process (such as a formula, workflow, task, or step) defined in the `formulation` section that executed or generated the attributed data.
```
jkowalleck, line 8954: per spec, `formulations` describe how a component came together.
> Describes how a component or service was manufactured or deployed. This is achieved through the use of formulas, workflows, tasks, and steps, which declare the precise steps to reproduce along with the observed formulas describing the steps which transpired in the manufacturing process.

per spec, `metadata.tools` list how the BOM came together.
> The tool(s) used in the creation, enrichment, and validation of the BOM.



since this feature intents to annotate/detail how the BOM  was made, I would expect to reference one of `metadata.tools` here, not a formulation. 
am i wrong?

jkowalleck: A solution i came up with, to make the current draft work: 
Change the spec, so that `formulation` could either point to a `$.components[]`/`$.services[]` (current spec) or the bom itself (new. Bom-link the `$.serialNumber` of the bom? Also the `$.version`? Or link to one of the `$.metadata.tools[]` ? Other ideas?)

jkowalleck: best would be to link to one of `$.metadata.tools[]` -- we do the same with evidence collection: https://cyclonedx.org/docs/1.6/json/#components_items_evidence_identity_oneOf_i0_items_tools
> The object in the BOM identified by its bom-ref. This is often a component or service but may be any object type supporting bom-refs. Tools used for analysis should already be defined in the BOM, either in the metadata/tools, components, or formulation.

stevespringett: Formulation description has been updated. Added support for expressions (JSONPath and XPath)

stevespringett:
The updated PR adds support for "expressions" supporting both JSONPath and XPath. The choice of using a pointer or an expression has been implemented in JSON and XML and explained in the protobuf.

The definition of formulation has been extended to capture its true purpose.

PR #661, opened 2025-07-17
CBOM 1.7: Update test cases & a few schema fixes/extensions

bhess:
Update schema test cases and fix validation issues for 1.7 CBOM.

This PR addresses two schema validation issues discovered while running ajv on the test cases:
- ikev2TransformTypes: The array items were missing type definitions. This has been corrected.
- certificateExtensions: The oneOf clause defined two objects (commonExtensions and customExtensions) with overlapping property names, which caused validation errors. The property names have been made unique within each object to resolve this.

Additionally, two new properties have been added to cryptoProperties.cipherSuites to support more precise TLS configurations: tlsGroups and tlsSignatureSchemes.


bhess:
It appears that the Java workflow is currently failing with the following error:

```
JsonSchemaVerificationTest.lambda$dynamicTestsWithCollection$2:102->isValid:117 » InvalidSchema Schema from 'http://cyclonedx.org/schema/cryptography-defs.schema.json' is not allowed to be loaded.
```

From what I can tell, this may not indicate a real issue with the schema itself. The error suggests that the validator is attempting to load the schema from a remote URL, which is currently unavailable or disallowed by the validator's configuration. Since http://cyclonedx.org/schema/cryptography-defs.schema.json is not yet publicly accessible, this could be expected behavior.

schema/bom-1.7.schema.json
```diff
@@ -5760,6 +5760,10 @@
                 "wpa",
                 "dtls",
                 "quic",
+                "eap-aka",
```
stevespringett, line 5763: Add `eap-aka-prime`

schema/bom-1.7.schema.json
```diff
@@ -5772,6 +5776,10 @@
                 "wpa": "Wi-Fi Protected Access",
                 "dtls": "Datagram Transport Layer Security",
                 "quic": "Quick UDP Internet Connections",
+                "eap-aka": "Extensible Authentication Protocol variant",
```
stevespringett, line 5780: Add eap-aka-prime

PR #682, opened 2025-09-08
Adds support for post-quantum ML-DSA algorithms

prabhu:
Fixes https://github.com/CycloneDX/specification/issues/674

With inspiration from the Node.js [PRs](https://github.com/nodejs/node/pull/59259) and [this](https://github.com/nodejs/node/pull/59461), this pull request adds three new post-quantum algorithms: "ML-DSA-44", "ML-DSA-65", and "ML-DSA-87" to better future-proof BOM signature validity. The hypothetical but real threat this addresses is that traditional signatures with RSA/ECDSA may become invalid when quantum computers break them.

Following alternative ideas could be considered:

1. URI-Based references

```"algorithm": "urn:nist.gov:fips204:ml-dsa-44"``` instead of simple upper case enum strings. This however is a significant breaking change to replace all existing identifiers in use.

2. Extend `signer` definition

Signature object support `signer` as one of the types. A new type called `postQuantumSigner` could be created to extend ` signer` to capture additional attributes related to PQ algorithms and also to support a broad range of PQ algorithms.

```
"properties": {
            "signers": {
              "type": "array",
              "title": "Signature",
              "description": "Unique top level property for Multiple Signatures. (multisignature)",
              "items": {"$ref": "#/definitions/signer"}
            }
          }
```

3. Enable additional properties

Currently, additional properties are disabled for signer. To increase flexibility, this could be enabled.

This PR proposes a simple enhancement to the enum, key type indicators, and validation, making this feature >= 1.7 only without any possibility for backports.

TBD:

Enhancing CBOM to support the corresponding OID values.

| Algorithm   | OID                          |
|-------------|------------------------------|
| ml-dsa-44   | 2.16.840.1.101.3.4.3.17      |
| ml-dsa-65   | 2.16.840.1.101.3.4.3.18      |
| ml-dsa-87   | 2.16.840.1.101.3.4.3.19      |

### DIFF
--- a/schema/jsf-0.82.schema.json
+++ b/schema/jsf-0.82.schema.json
@@ -3,7 +3,7 @@
   "$id": "http://cyclonedx.org/schema/jsf-0.82.schema.json",
   "type": "object",
   "title": "JSON Signature Format (JSF) standard",
-  "$comment" : "JSON Signature Format schema is published under the terms of the Apache License 2.0. JSF was developed by Anders Rundgren (anders.rundgren.net@gmail.com) as a part of the OpenKeyStore project. This schema supports the entirely of the JSF standard excluding 'extensions'.",
+  "$comment" : "JSON Signature Format schema is published under the terms of the Apache License 2.0. JSF was developed by Anders Rundgren (anders.rundgren.net@gmail.com) as a part of the OpenKeyStore project. This schema extends the JSF standard to include post-quantum algorithms (ML-DSA) following NIST FIPS 204, excluding 'extensions'.",
   "definitions": {
     "signature": {
       "type": "object",
@@ -15,7 +15,7 @@
             "signers": {
               "type": "array",
               "title": "Signature",
-              "description": "Unique top level property for Multiple Signatures. (multisignature)",
+              "description": "Unique top level property for Multiple Signatures. (multisignature). Can include both classical and post-quantum signatures.",
               "items": {"$ref": "#/definitions/signer"}
             }
           }
@@ -26,14 +26,14 @@
             "chain": {
               "type": "array",
               "title": "Signature",
-              "description": "Unique top level property for Signature Chains. (signaturechain)",
+              "description": "Unique top level property for Signature Chains. (signaturechain). Chain can include both classical and post-quantum signatures following NIST FIPS 204 standards.",
               "items": {"$ref": "#/definitions/signer"}
             }
           }
         },
         {
           "title": "Signature",
-          "description": "Unique top level property for simple signatures. (signaturecore)",
+          "description": "Unique top level property for simple signatures. (signaturecore). Supports both classical and post-quantum algorithms including ML-DSA variants.",
           "$ref": "#/definitions/signer"
         }
       ]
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: schema/jsf-0.82.schema.json [hunk 1]
```diff
@@ -15,7 +15,7 @@
             "signers": {
               "type": "array",
               "title": "Signature",
-              "description": "Unique top level property for Multiple Signatures. (multisignature)",
+              "description": "Unique top level property for Multiple Signatures. (multisignature). Can include both classical and post-quantum signatures.",
```
prabhu, line 18: Question: Ideally, this shouldn't be allowed, since it creates a weaker link. Perhaps this is needed to allow everyone to upgrade?

schema/jsf-0.82.schema.json
```diff
@@ -52,7 +52,7 @@
             {
               "type": "string",
               "title": "Algorithm",
-              "description": "Signature algorithm. The currently recognized JWA [RFC7518] and RFC8037 [RFC8037] asymmetric key algorithms. Note: Unlike RFC8037 [RFC8037] JSF requires explicit Ed* algorithm names instead of \"EdDSA\".",
+              "description": "Signature algorithm. The currently recognized JWA [RFC7518] and RFC8037 [RFC8037] asymmetric key algorithms, plus post-quantum algorithms. Note: Unlike RFC8037 [RFC8037] JSF requires explicit Ed* algorithm names instead of \"EdDSA\". Three post-quantum ML-DSA algorithms have been added following NIST FIPS 204 standardization.",
```
prabhu, line 55: Question: Should the description be enhanced to steer the users towards the new algorithms? Or should it remain this way and unopinionated?

schema/jsf-0.82.schema.json
```diff
@@ -81,49 +84,50 @@
...
-          "description": "The signature data. Note that the binary representation must follow the JWA [RFC7518] specifications."
+          "description": "The signature data encoded as base64url without padding. Encoding standards: (1) Classical algorithms (RS256, ES256, Ed25519, etc.): Follow JWA [RFC7518] specifications for signature serialization. (2) Post-quantum algorithms (ML-DSA-44, ML-DSA-65, ML-DSA-87): Follow NIST FIPS 204 standard for ML-DSA signature encoding."
         }
       }
     },
     "keyType": {
       "type": "string",
       "title": "Key type",
-      "description": "Key type indicator.",
+      "description": "Key type indicator. 'PQ' indicates post-quantum algorithms following NIST FIPS 204 standards, specifically ML-DSA algorithms.",
```
prabhu, line 120: Is this enough? Or should we expand the semantics for all key types?

schema/jsf-0.82.schema.json
```diff
@@ -81,49 +84,50 @@
...
+          "description": "Optional. Array holding the names of one or more application level properties that must be excluded from the signature process. Note that the \"excludes\" property itself, must also be excluded from the signature process. Since both the \"excludes\" property and the associated data it points to are unsigned, a conforming JSF implementation must provide options for specifying which properties to accept. This applies to both classical and post-quantum signatures.",
           "items": {
             "type": "string"
           }
         },
         "value": {
           "type": "string",
           "title": "Signature",
-          "description": "The signature data. Note that the binary representation must follow the JWA [RFC7518] specifications."
+          "description": "The signature data encoded as base64url without padding. Encoding standards: (1) Classical algorithms (RS256, ES256, Ed25519, etc.): Follow JWA [RFC7518] specifications for signature serialization. (2) Post-quantum algorithms (ML-DSA-44, ML-DSA-65, ML-DSA-87): Follow NIST FIPS 204 standard for ML-DSA signature encoding."
```
prabhu, line 113: `encoded as base64url without padding` - could this requirement be perceived as a breaking change for existing classical algorithms?

schema/jsf-0.82.schema.json
bhess: The semantics of multi classical/pq signatures will commonly be to require that both validate successfully, to avoid the weaker link.

Issue #719, opened 2025-11-12
[FEATURE]: Make VEX Ratings Actionable in CycloneDX Consumers


fahedouch:
### Describe the feature

Many CycloneDX VEX consumers (e.g `Dependency-Track`, `Trivy`..) consider only `analysis.state` (e.g., `not_affected`, `exploitable`, `resolved`) and ignore `vulnerabilities[].ratings[]` (e.g., CVSS, OWASP Risk Rating). Ratings carry essential exploitability and risk context and should be first-class inputs for prioritization.

Why this matters: ratings are especially important when [analysis.state](https://cyclonedx.org/docs/1.6/json/#vulnerabilities_items_analysis_state) is `exploitable` or `in_triage`. In these cases, ratings (e.g., OWASP Risk Rating, CVSS) provide the precision needed to assess exploitability and prioritize remediation. VEX is currently the most effective community mechanism to exchange this information.

Request: Provide clear guidance or a mechanism in CycloneDX so that `ratings` are consistently interpreted and used by scanners alongside `analysis.state`.

### Possible solutions

- Spec clarification: Normative language that consumers SHOULD/MUST consider `ratings` with `analysis.state`, with basic precedence guidance.
- Introduce new specification to exchange ratings: Vulnerability Rating Exchange (VRX)

### Additional context

Observed behavior: tools often ignore OWASP/CVSS entries in VEX `ratings`.

Example intent:
```json
{
  "vulnerabilities": [
    {
      "id": "DEBIAN-CVE-XXX",
      "ratings": [{ "method": "OWASP", "score": 5.0, "severity": "critical" }],
      "analysis": {
        "state": "in_triage",
        "detail": "Severity via OWASP Risk Rating (score 5.0 → critical)"
      }
    }
  ]
}
```

Questions
- Should `ratings` be normative inputs for prioritization in VEX consumers?
- Should we introduce a new specification for `ratings` exchange (VRX)?

PR #722, opened 2025-11-14
Update ratings descriptions in schema files for clarity on VEX usage

fahedouch:
I am translating @stevespringett 's [feedback](https://github.com/CycloneDX/specification/issues/719#issuecomment-3528968238) on the CycloneDX VEX specification into the code.

>  Should ratings be normative inputs for prioritization in VEX consumers?

_Yes, they should be. It is widely known that the NVD has historically overrated vulnerabilities (on purpose). So the ratings from the NVD and those from the manufactures are often different. CycloneDX can convey this information which can aid in prioritization._

fixes  https://github.com/CycloneDX/specification/issues/719

### DIFF
--- a/schema/bom-1.7.schema.json
+++ b/schema/bom-1.7.schema.json
@@ -2841,7 +2841,7 @@
         "ratings": {
           "type": "array",
           "title": "Ratings",
-          "description": "List of vulnerability ratings. Consumers SHOULD use ratings and SHOULD NOT ignore them; source ratings may differ and aid prioritization.",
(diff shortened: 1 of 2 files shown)


schema/bom-1.7.schema.json
```diff
@@ -2841,7 +2841,7 @@
         "ratings": {
           "type": "array",
           "title": "Ratings",
-          "description": "List of vulnerability ratings",
+          "description": "List of vulnerability ratings. Consumers SHOULD use ratings and SHOULD NOT ignore them; source ratings may differ and aid prioritization.",
```
fahedouch: Should I document how ratings should be interpreted and applied by consumers ?

PR #753, opened 2025-12-30
fix(registry): remove duplicate BLAKE2b-HMAC variant (#752)

### LINKED ISSUE #752: [Defect]: Duplicate BLAKE2b-HMAC variant in Cryptography Registry
## Describe the defect

**Where:**
schema/cryptography-defs.json

**What:**
The BLAKE2 family variants include a duplicate entry for:
- BLAKE2b-(160|256|384|512)-HMAC (primitive: mac)

**Impact:**
Duplicate variant patterns can cause ambiguity and inconsistent processing across consumers.

## Additional context

Data-quality cleanup only. No new algorithms or schema behavior changes.

Mehrn0ush:
As discussed in ticket #752, this PR fixes a data-quality issue in the Cryptography Registry.

Fixes #752

### Details
- Removes a duplicate `BLAKE2b-(160|256|384|512)-HMAC` variant entry (primitive: mac)

### Scope
- Cryptography Registry data only (`schema/cryptography-defs.json`)
- No schema or specification behavior changes

### DIFF
--- a/schema/cryptography-defs.json
+++ b/schema/cryptography-defs.json
@@ -122,6 +122,21 @@
           "primitive": "pke"
         }
       ]
+    },   
+    {
+      "family": "HPKE",
+      "standard": [
+        {
+          "name": "RFC9180",
+          "url": "https://doi.org/10.17487/RFC9180"
+        }
+      ],
+      "variant": [
+        {
+          "pattern": "HPKE[-(mode_base|mode_psk|mode_auth|mode_auth_psk)]-{kem}-{kdf}-{aead}",
+          "primitive": "pke"
+        }
+      ]
     },
     {
       "family": "MQV",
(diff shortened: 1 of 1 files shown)


stevespringett:
cc: @bhess

schema/cryptography-defs.json
```diff
@@ -933,10 +933,6 @@
         {
           "pattern": "BLAKE2s-(160|256)",
           "primitive": "hash"
-        },
-        {
-          "pattern": "BLAKE2b-(160|256|384|512)-HMAC",
```
bhess, line 977: The intention was to describe BLAKE2s with HMAC (instead of BLAKE2b), so this appears to be a typo that caused the duplicate. Could you please update the pattern to `BLAKE2s-(160|256|384|512)-HMAC` rather than removing the entry?

bhess (changes requested):
Thank you for highlighting this issue, @Mehrn0ush. Please see my inline comment, it seems to be a typo that caused the duplicate.

schema/cryptography-defs.json
Mehrn0ush: Thanks for the review, @bhess — agreed this was a typo.

I updated the entry to preserve both HMAC variants by correcting the duplicate to:
- `BLAKE2s-(160|256)-HMAC`

Note: BLAKE2s only defines 160/256-bit digests (whereas BLAKE2b supports up to 512), so I kept the size set aligned with the existing `BLAKE2s-(160|256)` variant.

PR updated (amended) accordingly.

bhess: Indeed, thanks!

schema/cryptography-defs.json
```diff
@@ -338,6 +338,16 @@
           "pattern": "AES[-(128|192|256)][-(GCM|CCM)][-{tagLength}][-{ivLength}]",
           "primitive": "ae"
         },
+        {
+          "standard": [
```
bhess, line 357: Please remove this entry to avoid merge conflicts with your other PR introducing AES-GCM-SIV

PR #774, opened 2026-01-13
Implements conditional subschemas based on component type

### LINKED ISSUE #638: CycloneDX 2.0 - Constrain Properties to Applicable Types
### Problem

In CycloneDX 1.x, several properties are loosely defined and may appear on types where they are semantically invalid. For example:

- `cryptoProperties` is intended only for cryptographic assets (e.g. cryptographic libraries, keys, algorithms), but is currently allowed on any `component`.
- `modelCard`, `evaluationMeasures`, and other AI/ML-specific metadata may appear on non-AI components.

This flexibility was useful during schema evolution, but introduces:
- Semantic ambiguity  
- Inconsistent SBOM usage  
- Misleading tool or ecosystem behavior  

---

### Goal for 2.0

Enforce **contextual constraints** on properties by:
- Restricting `cryptoProperties` to cryptographic asset types (e.g. `library`, `firmware`, `machine-learning-model`, etc.)
- Limiting AI/ML-specific fields (`modelCard`, `evaluationMeasures`, `datasets`, etc.) to components of `type: machine-learning-model`
- This ticket is not limited to `cryptoProperties` or `modelCard` and may be applicable to other properties in v1.6 or which will be introduced in v1.7.

---

### Proposal

Update the CycloneDX 2.0 schema to:
1. Apply `oneOf` or `if/then/else` constraints where schema logic permits
2. Scope optional fields to valid `component.type` or `service.category` values

--- 

As written, this is not a breaking change and the documentation in the current specification clearly states the purpose and when to use these properties. This ticket is for the enforcement of that behavior through minor changes in schema structure.

stevespringett:
Implements and closes #638

### DIFF
--- a/schema/2.0/model/cyclonedx-component-2.0.schema.json
+++ b/schema/2.0/model/cyclonedx-component-2.0.schema.json
@@ -250,20 +250,6 @@
           "title": "Release notes",
           "description": "Specifies release notes."
         },
-        "modelCard": {
-          "$ref": "cyclonedx-ai-modelcard-2.0.schema.json#/$defs/modelCard",
-          "title": "AI/ML Model Card"
-        },
-        "data": {
-          "type": "array",
-          "items": {"$ref": "#/$defs/componentData"},
-          "title": "Data",
-          "description": "This object SHOULD be specified for any component of type `data` and must not be specified for other component types."
-        },
-        "cryptoProperties": {
-          "$ref": "cyclonedx-cryptography-2.0.schema.json#/$defs/cryptoProperties",
-          "title": "Cryptographic Properties"
-        },
         "tags": {
           "$ref": "cyclonedx-common-2.0.schema.json#/$defs/tags",
           "title": "Tags"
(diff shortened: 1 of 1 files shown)


stevespringett:
@Mehrn0ush thank you for the thorough review.

> Right now the rule triggers on the presence of versionRange, and the then clause only constrains isExternal to be true if isExternal is present. Since isExternal is not required there, an object can still include versionRange while omitting isExternal, and it will pass validation.

I believe that is accurate and captures the current behavior in v1.7

> If the intent is “versionRange may only be used when isExternal is true” (as the versionRange field description suggests), a more direct pattern would be:

@jkowalleck I know you worked on this. What are your thoughts. I attempted to keep the existing logic in place - and I think I have.

> One small wording consistency nit: the Version Range rule description currently reads “versionRange must not be present when isExternal is false”, while the versionRange field description says “May only be used if isExternal is true.” These are closely related, but it might be worth aligning the phrasing so the intent is unambiguous (and easier to carry into spec text later).

I'm open to improving the wording. Suggestions.

As far as the design goes. This issue passes the public RFC comment period.

jkowalleck:
Your current approach makes the spec messy with all those constraints glued on the thing.
Better give them names and extend from a base, instead of using conditional restrictions. 

----


alternative approach: 

after 
- https://github.com/CycloneDX/specification/pull/982
- https://github.com/CycloneDX/specification/pull/975

the current `component` should be split in 
- `baseComponent` - core properties of a component
- `serviceComponent` - inherit from `baseComponent` and add service-specific properties and requirements
- `softwareComponent` - inherit from `baseComponent` and add {application,framework,library,...}-specific properties and requirements
- `hardwareComponent` - inherit from `baseComponent` and add service-specific properties and requirements
- `fileComponent` - inherit from `baseComponent` and add file-specific properties and requirements
- `dataComponent` - inherit from `baseComponent` and add data-specific properties and requirements
- ... and so on - for each component type 


after that split, the current `component` becomes a oneOf 
`[ { "$ref": "3$defs/serviceComponent" }, ... ]`


put this idea into a ticket: https://github.com/CycloneDX/specification/issues/1024

schema/2.0/model/cyclonedx-component-2.0.schema.json
```diff
@@ -282,20 +268,91 @@
...
           "not": {
             "required": ["version", "versionRange"]
           }
         },
         {
+          "title": "Version Range Requirement",
           "description": "Requirement: 'versionRange' must not be present when 'isExternal' is `false`.",
           "if": {
-            "properties": { "isExternal": { "const": false } }
+            "required": ["versionRange"]
```
jkowalleck: this looks like a fix that is not in the scope of this very ticket.

jkowalleck, line 281: intention: 
? Requirement: 'versionRange' must not be present when 'isExternal' is `false`.

before change:
if `isExternal` is set to false, then `versionRange` must not be present.
-or- only if `isExternal` is true, then `versionRange` may be set.


| isExternal | versionRange present | result/valid per rule | 
| ---------- | -------------------- | --------------------- |
| 0 | 0 | 1 |
| 0 | 1 | 0 |
| 1 | 0 | 1 |
| 1 | 1 | 1 |

----

after change: 
if `versionRange` is set, then `isExternal` must be true
-or- only if `isExternal` is true, then `versionRange` may be set.


| isExternal | versionRange present | result/valid per rule | 
| ---------- | -------------------- | --------------------- |
| 0 | 0 | 1 |
| 0 | 1 | 0 |
| 1 | 0 | 1 |
| 1 | 1 | 1 |



while technically the "before" and "after" work the same, i find the "bnefore" better since it is more on the actual textuial description>

Mehrn0ush: Agreed that "before" and "after" are logically equivalent as truth tables — that part isn't in question. My preference for "after" (`versionRange` present → `isExternal` must be true) is really about consistency: the `versionRange` field description already says "May only be used if `isExternal` is true," which is that same direction. Wording the rule description the other way just means a reader has to mentally contrapose one against the other to confirm they agree.

Both truth tables treat `isExternal` as strictly binary (true/false), but in the schema it can also be absent. The current `if` block keys on `"isExternal": {"const": false}`, which won't match a missing property — so an object with `versionRange` set and `isExternal` omitted entirely still passes validation today. That's the actual defect I was pointing at originally, not just the wording. Whichever phrasing we land on, I think the spec text and the schema logic both need to explicitly account for the "isExternal omitted" case, not just true/false.

jkowalleck:
there are some merge conflicts that need to be taken care of

PR #776, opened 2026-01-15
Refactors role-specific properties into entities

### LINKED ISSUE #718: [REFACTOR]: Role-based organizationalEntities (manufacturers, suppliers, etc) into extensible objects
A lot of discussion in the HBOM working group about different types of roles an entity may play, and which need to be captured in an HBOM.

Based on these discussions, the idea is to replace the dedicated "manufacturer", "supplier" and similar properties that take on a role, with a generic "entities" property which supports both organizationalEntity and organizationalContact along with a "role" property.

Much like the lifecycles objects, you can use the enums or can define your own.

stevespringett:
Implements and closes #718


schema/2.0/model/cyclonedx-common-2.0.schema.json
```diff
@@ -407,6 +407,164 @@
     },
     "organizationalEntityOrContact": {
 
+    },
+    "entity": {
+      "type": "object",
+      "title": "Entity",
```
jkowalleck: "entity" for id, and "Entity" for title - a much too broad term.
please use a more narrow term.

background: every item of a collection is an entity. every banana is an entity of fruit. ...

schema/2.0/model/cyclonedx-common-2.0.schema.json
```diff
@@ -407,6 +407,164 @@
...
+      "oneOf": [
+        {
+          "required": ["person"]
+        },
+        {
+          "required": ["organization"]
+        }
+      ]
+    },
+    "role": {
```
jkowalleck, line 458: "role" for id, and "Role" for title - a much too broad term.
please use a more narrow term. 

i mean, if we ever have roles for services or something, you would call them "ServiceRoles", right? 
Better not use these broad and general terms for things that are pretty narrow in their scope, this prevents extensions in the future and might lead to confusion.

jkowalleck (changes requested):

schema/2.0/model/cyclonedx-common-2.0.schema.json
```diff
@@ -407,6 +407,164 @@
...
+        "roles": {
+          "type": "array",
+          "title": "Roles",
+          "description": "The role(s) that the entity fulfils. At least one role shall be specified. Multiple roles may be assigned to represent the various capacities in which the entity operates.",
+          "minItems": 1,
+          "items": {
+            "$ref": "#/$defs/role"
+          }
+        },
+        "priority": {
```
stevespringett: The "priority" label may be confusing. Perhaps "sourcingOrder" or just "order"

schema/2.0/model/cyclonedx-common-2.0.schema.json
```diff
@@ -407,6 +407,164 @@
...
+            "signatory",
+            "committer",
+            "auditor",
+            "integrator",
+            "operator",
+            "qualityControl",
+            "securityContact",
+            "supportContact",
+            "legalContact"
+          ],
```
stevespringett, line 510: Add additional roles for logistics. Freight forwarder for example.

stevespringett: Confirm with @bhess that the CA is captured in CBOM. May want to add certificateAuthority as a role though.

schema/2.0/model/cyclonedx-common-2.0.schema.json
stevespringett: Title changed

schema/2.0/model/cyclonedx-common-2.0.schema.json
stevespringett: priority changed to "order"

schema/2.0/model/cyclonedx-common-2.0.schema.json
stevespringett: Confirmed and added.

Mehrn0ush:
Hi @stevespringett  — quick follow-ups (low-impact), in case you think they’re worth folding in:

1) Prevent enum/meta drift
Now that the role taxonomy is growing, it might be worth adding a lightweight consistency check (either a CI step or a small script that CI can call) to ensure `role.enum` and `meta:enum` stay in sync—i.e., every `enum` value has a corresponding `meta:enum` description, and there are no extra `meta:enum` keys not present in `enum`. This helps prevent accidental omissions and keeps schema-driven docs/tooling consistent as the list evolves.

2) Minor hardening for custom roles
Consider adding `minLength: 1` for `custom role.name` (and optionally a short note discouraging whitespace-only values). In practice, empty/blank role names can slip through generators and produce confusing downstream diffs; a minimal constraint improves data quality and cross-tool consistency without changing the model.

3) Naming/wording consistency for `order`
Tiny consistency nit (similar to the recent title/description polish in this PR): the property is now `order`, but the `title`/wording still uses “Priority” in a few places. Since schema-driven UI/docs often surface `title` as the field label, this mixed terminology can confuse implementers (and may leak into spec-text wording later). If the intent is preference ordering, aligning the title/description terminology with `order` (or “preference order”) would make the semantics and migration clearer.

stevespringett:
@Mehrn0ush 
> Prevent enum/meta drift

Good point. We've created our own linter for CycloneDX 2.0, which I believe handles this case. We still need to hook it up into a build process, but the framework should be there.
https://github.com/CycloneDX/specification/blob/2.0-dev/tools/src/main/js/linter/checks/enum-value-formatting.check.js

> Minor hardening for custom roles

I don't think we have a linter check for this today and I know we've had an issue or two where this has come up, allowing someone to simply use `""` and it satisfies the required check, which defeats the purpose. The specific occurrence of this has now been corrected.

> Naming/wording consistency for order

Thanks for catching this. This has been corrected.

schema/2.0/model/cyclonedx-common-2.0.schema.json
```diff
@@ -407,6 +407,198 @@
     },
     "organizationalEntityOrContact": {
 
+    },
+    "entity": {
```
jkowalleck, line 411: i still dont like the wording "entity".
an entity of what?  better name: "organizationOrPerson"

schema/2.0/model/cyclonedx-common-2.0.schema.json
jkowalleck: then the name should also change to "organizationOrPersonRole"

jkowalleck (changes requested):

schema/2.0/model/cyclonedx-common-2.0.schema.json
Mehrn0ush: The term `entity` is used here as a domain-level abstraction representing an identifiable actor within CycloneDX.

Although the current `oneOf` constrains implementations to organization/person, this reflects the present schema scope rather than the conceptual boundary of the model.

Encoding concrete types directly into the primitive name (e.g. `organizationOrPerson`) tightly couples the abstraction to today’s variants and reduces extensibility for future actor classes (such as services, signing authorities, or automated agents).

Keeping `entity` generic aligns with standard domain modeling practice: define a stable abstract actor, then specialize via schema composition. This avoids introducing parallel actor constructs as the specification evolves.

Mehrn0ush: `role` models the contextual function an entity assumes (e.g. supplier, manufacturer, operator), not an intrinsic attribute of organization/person.

The structural relationship is:

  entity → roles[]

This follows established actor/role modeling patterns (RBAC / domain-driven design), where actors remain generic and roles express situational responsibility.

Renaming to `organizationOrPersonRole` couples role semantics to specific actor variants and limits reuse if additional entity types are introduced later.

Keeping `role` generic preserves a clean separation between actor identity and functional responsibility, and prevents fragmentation of role taxonomies across future entity categories.

stevespringett: > Renaming to organizationOrPersonRole couples role semantics to specific actor variants and limits reuse if additional entity types are introduced later.

That is precisely what the term "entity" was chosen. We've had a number of discussions in the threat modeling working group about different types of entities in the future.

> This follows established actor/role modeling patterns (RBAC / domain-driven design), where actors remain generic and roles express situational responsibility.

Perhaps change the name from "entity" to "actor"? We have this same concept in the threat model schema I believe, but it really needs to be a common object as its going to be reused everywhere.

stevespringett:
Closing as rejected.

PR #777, opened 2026-01-17
schema: refactor data/threat models and refresh bundles

P3tra-WP:
Title:
Refactor data/threat modeling schemas and regenerate bundles

Description:
This PR updates the data and threat modeling schemas to improve consistency, reuse, and explicit linkage across models, and regenerates bundled schemas.

What changed

Data classification refactor:
Moved detailed dataClassification into the shared data model.
dataClassification now supports: enum string, custom string, or detailed object.
Detailed object uses dataCategory for dataTypes.
Data objects and categories:
dataObject and dataCategory moved to cyclonedx-data-2.0.schema.json.
dataSet.dataObjects now references shared dataObject definitions.
Flow metadata consolidation:
Removed flow.dataFormat and flow.classification; flows reference dataObjects for these details.
Threat model enhancements:
Added vulnerabilityRef on threatScenario to link threats to vulnerabilities.
Added ibmRiskAtlas reference object.
CAPEC references already supported via attackPattern / attackPatternReference.
Risk model fixups:
Added missing likelihoodFactor definition.
Blueprint schema fixups:
Moved actor into $defs and added accessControlType alias to authorizationType.
Enforced dataObject classification via oneOf (inline vs ref).
Bundled outputs regenerated:
cyclonedx-2.0-bundled.schema.json
cyclonedx-2.0-bundled.min.schema.json
cyclonedx-api-2.0-bundled.schema.json
cyclonedx-api-2.0-bundled.min.schema.json
Notes

Bundler warns about missing 2020-12 meta-schema in AJV (existing behavior).
Testing

Bundled schemas regenerated via bundle-schemas.js.


schema/2.0/model/cyclonedx-data-2.0.schema.json
```diff
@@ -6,9 +6,333 @@
...
+          ],
+          "description": "Data classification level",
+          "meta:enum": {
+            "public": "Public information that can be freely shared",
+            "internal": "Internal use only within the organization",
+            "confidential": "Confidential information requiring protection",
+            "restricted": "Highly restricted information with limited access and special protection measures"
+          }
+        },
+        {
```
stevespringett: Remove the second object.

schema/2.0/model/cyclonedx-data-2.0.schema.json
```diff
@@ -6,9 +6,333 @@
...
+              "restricted"
+            ]
+          }
+        },
+        {
+          "type": "object",
+          "description": "Detailed data classification metadata.",
+          "additionalProperties": false,
+          "properties": {
+            "level": {
```
stevespringett, line 31: This is where the oneOf should go IMO. So that existing levels or custom levels can be specified and still take advantage of what the object provides.

schema/2.0/model/cyclonedx-data-2.0.schema.json
```diff
@@ -6,9 +6,333 @@
...
+                    "pseudonymize": "Pseudonymize identifiers and keep mapping separately",
+                    "mask": "Mask or obfuscate sensitive fields",
+                    "tokenize": "Replace sensitive values with tokens",
+                    "minimize": "Collect and retain only necessary data",
+                    "segregate": "Segregate or isolate sensitive data",
+                    "audit": "Enable audit logging for access/use",
+                    "custom": "Custom handling requirement"
+                  }
+                },
+                {
```
stevespringett, line 103: The second option should not be a string, rather an object with a name and description.

schema/2.0/model/cyclonedx-data-2.0.schema.json
```diff
@@ -6,9 +6,333 @@
...
+                      "minimize",
+                      "segregate",
+                      "audit",
+                      "custom"
+                    ]
+                  }
+                }
+              ]
+            },
+            "retention": {
```
stevespringett, line 122: The retention assumes persisted data. IMO, there is a need to also handle transient data such as PHI in a short-lived JWT that expires in an hour.

stevespringett: Petra mentioned TTL, which could be used alongside retention. We could also think about a simple integer field that is defined with a specific time-period. ie. nanoseconds, hours, etc.

schema/2.0/model/cyclonedx-data-2.0.schema.json
```diff
@@ -6,9 +6,333 @@
...
+                },
+                {
+                  "type": "string",
+                  "pattern": "^[0-9]+$",
+                  "description": "Retention period in months (numeric string)"
+                }
+              ]
+            },
+            "disposal": {
+              "oneOf": [
```
stevespringett: Change disposal to use the behavior taxonomy. See Steve for a prototype of this in action.

schema/2.0/model/cyclonedx-data-2.0.schema.json
```diff
@@ -6,9 +6,333 @@
...
+                      "gdpr",
+                      "ccpa",
+                      "hipaa",
+                      "pciDss",
+                      "glba",
+                      "sox",
+                      "ferpa",
+                      "coppa",
+                      "iso27001",
+                      "nist",
```
stevespringett, line 182: Perhaps change this to compliance requirements. nist is a standards body and they have multiple standareds that may be required such as FedRAMP, SSDF, CSF, and 800-53r5.

Ideally, referencing CDXA would be ideal. definitions/standards/
Need to investigate how to reuse.

schema/2.0/model/cyclonedx-data-2.0.schema.json
```diff
@@ -6,9 +6,333 @@
...
-      "description": "Data classification tags data according to its type, sensitivity, and value if altered, stolen, or destroyed."
+      "description": "Data classification tags data according to its level or a custom classification label."
+    },
+    "dataCategory": {
+      "type": "string",
+      "enum": [
+        "pii",
+        "phi",
+        "pci",
+        "financial",
```
stevespringett, line 234: PIFI from public financial information

schema/2.0/model/cyclonedx-data-2.0.schema.json
```diff
@@ -6,9 +6,333 @@
...
+        "intellectualProperty",
+        "telemetry",
+        "logs",
+        "credentials",
+        "safety",
+        "operational",
+        "custom"
```
stevespringett, line 248: biometric, etc. should be supported.

schema/2.0/model/cyclonedx-data-2.0.schema.json
```diff
@@ -6,9 +6,333 @@
...
+          "description": "Reference to a shared classification definition elsewhere in the model"
+        },
+        "categories": {
+          "type": "array",
+          "items": {
+            "$ref": "#/$defs/dataCategory"
+          },
+          "description": "Categories of data (PII, PHI, PCI, etc.)"
+        },
+        "schema": {
```
stevespringett, line 308: for schema and format, work with Steve. We're doing something similar with exterenalReferences. Related to #185.

schema/2.0/model/cyclonedx-threat-2.0.schema.json
```diff
@@ -36,6 +40,9 @@
             "weakness": {
               "$ref": "#/$defs/weaknessReference"
             },
+            "ibmRiskAtlas": {
```
stevespringett: rename this so that it is not IBM specific.

schema/2.0/model/cyclonedx-threat-2.0.schema.json
```diff
@@ -348,6 +355,21 @@
         }
       }
     },
+    "ibmRiskAtlasReference": {
```
stevespringett: Rename -  make it more universal

schema/2.0/model/cyclonedx-threat-2.0.schema.json
```diff
@@ -348,6 +355,21 @@
...
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
```
stevespringett: add a new field  to represent "IBM Risk Atlas" and any other custom taxonomy.

stevespringett: Take a look at the vulnerabiity schema which supports CWE and multiple ratings, including CSVV, SSVC, and OWASP Risk Rating.

stevespringett: In the case of ASVS and similar, these can also be handled in CDXA.

stevespringett (changes requested):
Comments inline

schema/2.0/model/cyclonedx-data-2.0.schema.json
Mehrn0ush: Retention here is modeled as “months” which implies persisted storage. To cover transient data (JWT/session tokens), I suggest adding a separate duration field (e.g., ttl or lifetime) rather than overloading retention.
Proposed options:
	•	ISO 8601 duration: ttl: "PT1H" / ttl: "P7D" (simple, language-agnostic)
	•	or { value: 3600, unit: "seconds" } (explicit units)
This would let implementers represent “persisted retention” and “transient lifetime” simultaneously without ambiguity.

schema/2.0/model/cyclonedx-data-2.0.schema.json
Mehrn0ush: I think we can make this both implementer-friendly and schema-consistent by moving the oneOf to dataClassification.level inside the detailed object.
Proposed shape:
	level: oneOf [ {enum: public|internal|confidential|restricted}, {type: object, required: [name], properties: {name, description}} ]
This keeps the benefits of the rich object (handling/retention/disposal/regulations) while still supporting custom classification labels without losing metadata.

schema/2.0/model/cyclonedx-data-2.0.schema.json
```diff
@@ -23,38 +23,47 @@
...
-              "restricted"
-            ]
-          }
-        },
         {
           "type": "object",
           "description": "Detailed data classification metadata.",
           "additionalProperties": false,
           "properties": {
             "level": {
```
Mehrn0ush, line 31: In the dataClassification object branch, level is defined but not required—so an empty object {} would validate. Would it make sense to require level ("required": ["level"]) given this branch is intended for detailed metadata? The simple string form already covers cases without metadata, so this would align the object form with its intent.

schema/2.0/model/cyclonedx-data-2.0.schema.json
Mehrn0ush, line 230: The enum expansion is great, but dataCategory will likely keep growing over time. To avoid repeated enum churn while keeping strong interop, would you consider making dataCategory extensible via oneOf (standard enum string OR a custom { name, description } object)? This preserves machine-comparable common values while allowing stable vendor/custom categories without further schema changes.

Issue #756, opened 2025-12-31
[FEATURE]: Add Argon2 to Cryptography Registry


beatquantum:
## Problem Statement
Argon2 (RFC 9106 standard, winner of 2015 Password Hashing Competition) 
is not currently in the CycloneDX Cryptography Registry despite widespread 
adoption and OWASP recommendation.

## Proposed Pattern
Argon2(id|i|d)[-{memoryMiB}][-{iterations}][-{parallelism}]

Valid examples:
- Argon2id (variant only)
- Argon2id-15-1-4 (RFC 9106 interactive defaults)
- Argon2id-64-2-1 (medium security)
- Argon2id-256-4-2 (high security)

## Illustration of Compatibility (json)

{
  "type": "cryptographic-asset",
  "name": "Argon2id-64-2-1",
  "cryptoProperties": {
    "assetType": "algorithm",
    "algorithmProperties": {
      "primitive": "password-hashing",
      "algorithmFamily": "Argon2",
      "parameterSetIdentifier": "id-64-2-1"
    }
  }
}

## Justification
1. Parametric algorithms are already in registry
2. Parameter variation critically impacts security
3. Enables policy enforcement and CBOM risk assessment
4. RFC 9106 is authoritative standard

## Standards Reference
RFC 9106: Argon2 Memory-Hard Function for Password Hashing

stevespringett:
cc: @bhess

meghsaha-ibm:
The primitive for the variant Argon2(id|i|d)[-{memoryKiB}][-{passes}][-{parallelism}] should be **kdf**. Also, the variable {dkLen} may be considered.

PR #793, opened 2026-01-25
Add Argon2 (RFC 9106) to Cryptography Registry

Mehrn0ush:
This PR adds Argon2 (RFC 9106) to `schema/cryptography-defs.json`.

## Notes on parameter naming vs Issue #756
Issue #756 proposes: `Argon2(id|i|d)[-{memoryMiB}][-{iterations}][-{parallelism}]`.

RFC 9106 defines:
- `m` as **memory size in kibibytes (KiB)**,
- `t` as **number of passes**,
- `p` as **degree of parallelism**.

Accordingly, this PR uses:
`Argon2(id|i|d)[-{memoryKiB}][-{passes}][-{parallelism}]`

If the community prefers the Issue's naming for consistency with existing conventions, I can update the pattern accordingly.

Refs: #756

### DIFF
--- a/schema/cryptography-defs.json
+++ b/schema/cryptography-defs.json
@@ -1367,7 +1367,7 @@
       ],
       "variant": [
         {
-          "pattern": "Argon2(id|i|d)[-{memoryKiB}][-{passes}][-{parallelism}][-{saltLength}][-{secretLength}][-{associatedDataLength}][-{tagLength}]",
+          "pattern": "Argon2(id|i|d)[-{memoryKiB}][-{passes}][-{parallelism}][-{saltLenBytes}][-{secretLenBytes}][-{associatedDataLenBytes}][-{tagLenBytes}]",
           "primitive": "kdf"
         }
       ]


meghsaha-ibm:
The primitive for the variant Argon2(id|i|d)[-{memoryKiB}][-{passes}][-{parallelism}] should be **kdf**. Also, the variable {dkLen} may be considered.

Mehrn0ush:
I noticed schema/cryptography-defs.schema.json did not end with a trailing newline in the current base.
My earlier update inadvertently introduced an EOF newline change alongside the intended enum update.
This follow-up commit keeps the file’s original EOF formatting (no trailing newline) and only adds "Argon2" to algorithmFamiliesEnum, so the schema enum remains consistent with cryptography-defs.json without unrelated formatting noise.

Mehrn0ush:
@bhess  Thanks for the pointer — appreciated. 

I’ve updated the Argon2 variant pattern to include salt, secret value K, associated data, and tag lengths per RFC 9106 §3.1, and renamed the fields to *LenBytes to make the byte-based units explicit.

Added in a follow-up commit.

Let me know if you spot anything else worth capturing for completeness.

PR #812, opened 2026-02-05
Add ShangMi (SM2/SM3/SM4/SM9) algorithm families

### LINKED ISSUE #811: [FEATURE]: Add ShangMi (SM2/SM3/SM4/SM9) algorithm families to cryptography definitions
## Describe the feature

ShangMi (SM) cryptographic algorithms are mandated in multiple regulated environments in China, including TLS deployments, PKI infrastructures, and government systems.

Currently, CycloneDX does not define SM2, SM3, SM4, or SM9 in `cryptography-defs.json`. This prevents SBOM producers from accurately expressing ShangMi-based cryptographic implementations, limiting interoperability for software distributed into these environments.

Adding these families enables CycloneDX BOMs to represent widely deployed cryptographic stacks used in Chinese compliance contexts.

---

## Possible solutions

Add the following algorithm families to `cryptography-defs.json` and `algorithmFamiliesEnum`:

- **SM2** — signature / public-key encryption / key agreement  
- **SM3** — cryptographic hash  
- **SM4** — block cipher  
- **SM9** — identity-based cryptography

References are provided using:

- ISO standards (SM2/SM3/SM4)
- RFC 8998 (TLS integration of ShangMi algorithms)
- Public GM/T mirrors for SM9

A PR implementing this proposal is prepared.

---

## Alternatives

If preferred, SM9 could initially be represented as a single composite family (`primitive: other`) and further refined later based on usage patterns.

---

## Additional context

RFC 8998 documents production usage of SM2/SM3/SM4 in TLS 1.3 deployments.  
SM4 is also standardized internationally via ISO/IEC 18033-3.

This addition aligns CycloneDX cryptography definitions with existing national and international cryptographic ecosystems.

Mehrn0ush:
Fixes #811

Adds SM2/SM3/SM4/SM9 algorithm families to the CycloneDX cryptography registry and updates algorithmFamiliesEnum accordingly.

Note: some official ShangMi specification mirrors appear region-blocked or unreliable from multiple locations.
To avoid fragile links in the registry, this PR relies on globally accessible references (RFC Editor + ISO)

### DIFF
--- a/schema/cryptography-defs.json
+++ b/schema/cryptography-defs.json
@@ -297,7 +297,7 @@
           "primitive": "mac"
         },
         {
-          "pattern": "HMACXOF(128|256)",
+          "pattern": "KMACXOF(128|256)",
           "primitive": "mac"
         },
         {


schema/cryptography-defs.json
```diff
@@ -827,6 +827,74 @@
...
+      "standard": [
+        {
+          "name": "ISO/IEC18033-3:2010/Amd1:2021",
+          "url": "https://www.iso.org/standard/81564.html"
+        }
+      ],
+      "variant": [
+        {
+          "pattern": "SM4[-(ECB|CBC|CFB|OFB|CTR|XTS)][-{padding}][-{ivlen}]",
+          "primitive": "block-cipher"
```
bhess, line 914: the "ae" modes seem to be missing (e.g., CCM, GCM..)

schema/cryptography-defs.json
```diff
@@ -827,6 +827,74 @@
...
+        }
+      ]
+    },
+    {
+      "family": "SM9",
+      "standard": [
+        {
+          "name": "RFC8998",
```
bhess: SM9 seems to be not specified in RFC8998. See https://en.wikipedia.org/wiki/SM9_(cryptography_standard) for the references.

schema/cryptography-defs.json
```diff
@@ -827,6 +827,74 @@
...
+          "url": "https://www.rfc-editor.org/rfc/rfc8998.html"
+        }
+      ],
+      "variant": [
+        {
+          "pattern": "SM9",
```
bhess: rather than a single "other" primitive, it might be worth splitting this up to different variants.
I didn't look at the standard in the detail, but see https://en.wikipedia.org/wiki/SM9_(cryptography_standard), there are different primitives:
- signature
- keyagree/wrap
- kem

schema/cryptography-defs.json
Mehrn0ush: Thank you for the note — you’re absolutely right. I’ve added an explicit ae variant for SM4 covering GCM and CCM (pattern: SM4-(GCM|CCM)[-{tagLength}][-{ivLength}]), and kept the existing block-mode variant (ECB/CBC/CFB/OFB/CTR/XTS) under block-cipher. This mirrors the registry’s existing separation of block modes vs authenticated-encryption modes.

schema/cryptography-defs.json
Mehrn0ush: Thanks for pointing that out. You’re correct — SM9 is not specified in RFC8998. I updated the SM9 standard reference accordingly to GM/T0044.1-2016 and switched the URL to the corresponding GM/T document.

schema/cryptography-defs.json
Mehrn0ush: Appreciate the suggestion —  I replaced that entry using primitive other with multiple variants split by primitive to reflect the distinct SM9 functionalities:

signature via SM9-(SIG|SIGNATURE)

key-agree via SM9-(KEX|KEYEXCHANGE|KEY-EXCHANGE|KEYAGREE|KEYAGREEMENT|KEY-AGREE|KEY-AGREEMENT)

kem via SM9-(KEM|KEYENCAPSULATION|KEY-ENCAPSULATION)

pke via SM9-(ENC|ENCRYPTION|PKE|PUBLICKEY-ENCRYPTION|PUBLIC-KEY-ENCRYPTION)

If you’d prefer limiting SM9 strictly to signature / key agreement / KEM (as commonly summarized), I’m happy to drop the pke variant — I included it for completeness in case implementations label SM9 encryption explicitly.

Mehrn0ush:
Made SM2’s -256 suffix optional to accept both SM2 and SM2-256, following the registry’s [] convention. Updated patterns accordingly.

bhess (approved):
Thanks for the updates, LGTM!

PR #845, opened 2026-02-24
Cryptography: add unwrap, gmac, cmac

### LINKED ISSUE #780: [FEATURE]: Support for additional cipher modes in cryptoProperties.algorithmProperties.mode
Currently, the enumeration for cryptoProperties.algorithmProperties.mode does not include the following modes.

- GMAC
- CMAC

Could you please add the relevant items to the list of allowed values for modes. These modes are specified in variants of AES in https://cyclonedx.org/registry/cryptography/.

### LINKED ISSUE #794: [FEATURE]: Support for unwrap in cryptoProperties.algorithmProperties.cryptoFunctions
#769 added **wrap** to the list of cryptoProperties.algorithmProperties.cryptoFunctions but there is no item representing **unwrap**. Could you please add a relevant item to the list of allowed values for cryptoFunctions.

bhess:
Adds GMAC/CMAC to modes
Fixes #780 

Adds unwrap to cryptoFunctions
Fixes #794


schema/2.0/model/cyclonedx-cryptography-2.0.schema.json
```diff
@@ -274,6 +276,8 @@
                 "kwp": "AES Key Wrap with Padding (RFC 5649).",
                 "cts": "Ciphertext Stealing mode.",
                 "xts": "XEX Tweaked-codebook with Stealing (disk encryption).",
+                "gmac": "Galois message authentication code",
+                "cmac": "Cipher-based message authentication code",
```
Mehrn0ush: I just want to make sure I’m interpreting the modeling correctly.
algorithmProperties.mode is described as a block-cipher mode of operation, whereas gmac and cmac are typically treated as MAC constructions.

Are we intentionally allowing them here as AES variant suffixes (i.e., registry-aligned usage)? If so, would it make sense to slightly broaden the mode description to reflect that?

Alternatively, should these be modeled under primitive: mac (with cryptoFunctions: tag) and keep mode strictly for block-cipher modes?

Mehrn0ush: the meta:enum descriptions for gmac and cmac don’t currently end with a period, while most other entries do. Would you like to align those for consistency?

jvdsn: Tiny change but maybe capitalize it to be consistent with the other descriptions?

bhess: > I just want to make sure I’m interpreting the modeling correctly. algorithmProperties.mode is described as a block-cipher mode of operation, whereas gmac and cmac are typically treated as MAC constructions.
> 
> Are we intentionally allowing them here as AES variant suffixes (i.e., registry-aligned usage)? If so, would it make sense to slightly broaden the mode description to reflect that?
> 
> Alternatively, should these be modeled under primitive: mac (with cryptoFunctions: tag) and keep mode strictly for block-cipher modes?

I initially hesitated to categorize CMAC and GMAC as 'modes of operation' in the traditional sense. However, NIST SP 800-38D gave me an argument since it states: `GCM and GMAC are modes of operation for an underlying approved symmetric key block cipher.`

The mode property has become slightly redundant with the naming patterns in 1.7, so I would recommend to give the patterns the priority when creating CBOMs to express the algorithm details.

Mehrn0ush: That makes sense — thanks.

bhess: Agreed, updated accordingly.

PR #868, opened 2026-03-11
Add RSA-X931

### LINKED ISSUE #867: [FEATURE]: RSA X9.31 padding in cryptography registry
## Describe the feature

Currently the cryptography registry only has families for RSAES-OAEP, RSAES-PKCS1, RSASSA-PKCS1, and RSASSA-PSS. In https://github.com/CycloneDX/specification/pull/841/, X9.31 padding was added to the core spec, so it should also be in the cryptography registry.

## Additional context

I'm wondering if there should be a generic / catch-all "RSA" family, considering that there is also an "unknown" padding type in the core spec.

jvdsn:
fixes #867 

However, I was unable to find a good link for the standard yet, so this remains draft.

### DIFF
--- a/.github/workflows/build_docs.yml
+++ b/.github/workflows/build_docs.yml
@@ -21,10 +21,10 @@ jobs:
     steps:
     - name: Checkout
       # see https://github.com/actions/checkout
-      uses: actions/checkout@v6
+      uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
     - name: Set up JDK
       # see https://github.com/actions/setup-java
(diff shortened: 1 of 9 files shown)


jvdsn:
Could we reference https://standards.globalspec.com/std/1955293/ansi-x9-31 ? This is similar to what is done for X9.62: https://standards.globalspec.com/std/1955141/ansi-x9-62

bhess:
> Could we reference https://standards.globalspec.com/std/1955293/ansi-x9-31 ? This is similar to what is done for X9.62: https://standards.globalspec.com/std/1955141/ansi-x9-62

Sounds good to me.

PR #931, opened 2026-05-15
Add Crypto outputLength property to schema

### LINKED ISSUE #799: [Defect]: Missing variable {dkLen} in HKDF[-{hashFunction}] variant of the algorithm family HKDF
## Describe the defect

Missing variable {dkLen} in HKDF[-{hashFunction}] variant of the algorithm family HKDF.

## Additional context

The issue is data-quality / naming defect in the Cryptography Registry and can be fixed without changing schema behavior or introducing new algorithms.

jvdsn:
Fixes #799

Adds the outputLength property that indicates the output length of the cryptographic algorithm. This is not restricted to key derivation functions, but could also be used for MAC tags, XOF output lengths, and more.


jvdsn:
@meghsaha-ibm

bhess:
Thanks for the addition @jvdsn, that's useful! I had two questions regarding edge cases:
- For some algorithms, there are multiple outputs with different lengths. For example, KEM operations: encaps produces a ciphertext and shared secret, decaps produces a shared secret, and key generation produces a public/private key pair. Would it make sense to model outputs as a list of named values (e.g., { name: "public-key", length: X }) to avoid ambiguity when multiple outputs exist?
- Some algorithms have variable-length outputs (e.g., XOFs). Should this field support a range or otherwise indicate variable/unbounded length instead of a fixed value?

stevespringett:
Perhaps use oneOf to account for either an integer which would represent a single output length, or an object which would have min and max properties.

Refer to: https://json-schema.org/understanding-json-schema/reference/combining#oneOf

This would look something like:
```
"oneOf": [
    { "type": "integer" },
    { 
      "type": "object", 
      "properties": {
        "min": { ... }
        "max": { ... }
      }
    }
  ]
 ```

**Comment by jvdsn** [comment]
@stevespringett apologies for the delay. I have updated the PR now. Let me know what you think. One open question: do we want to enforce that both min and max are required (as they are currently)? Or should we allow one of them to be omitted?

**Comment by jvdsn** [comment]
@stevespringett I fixed the DCO

**Comment by jkowalleck** [comment]
RFC notice sent on July 23, 2026

* https://groups.io/g/CycloneDX/topic/120410973
* https://cyclonedx.slack.com/archives/CVA0G10FN/p1784815784761599

Public RFC period ends August 20, 2026

File: schema/2.0/model/cyclonedx-cryptography-2.0.schema.json [hunk 1]
```diff
@@ -325,6 +321,38 @@
...
+              "title": "Output length",
+              "description": "The output length provided by the cryptographic algorithm (in bits). For example: XOF output length, MAC tag length, derived key length for KDF.",
+              "oneOf": [
+                {
+                  "type": "integer",
+                  "minimum": 0
+                },
+                {
+                  "type": "object",
+                  "required": [
```
jkowalleck: are both required, always? 
not have a atleast one of them required?

jvdsn: Currently both are required for simplicity. If there's any strong opinions one way or another, I can change it.

jkowalleck: I'd love to see it as "at least one of them properties is required"

jkowalleck: this would be 
```patch
-                   "required": [
-                     "min",
-                     "max"
-                   ],
+                   "anyOf": [
+                     { "required": ["min"] },
+                     { "required": ["max"] },
+                   ]
```

jvdsn: @jkowalleck implemented now, please take a look.

jkowalleck (approved):

jkowalleck:
we 'd love to see some examples.
they could be added in the folder `tools/src/test/resources/2.0`

jkowalleck (changes requested):
could you add valid and invalid tests cases in https://github.com/CycloneDX/specification/tree/2.0-dev/tools/src/test/resources/2.0

jvdsn:
@jkowalleck added now, what do you think?

jkowalleck (approved):
generally approved from my side.

~~still missing test cases.~~

schema/2.0/model/cyclonedx-cryptography-2.0.schema.json
```diff
@@ -325,6 +325,46 @@
...
+                  "type": "integer",
+                  "minimum": 0
+                },
+                {
+                  "type": "object",
+                  "anyOf": [
+                    {
+                      "required": [
+                        "min"
+                      ]
```
jkowalleck:
```suggestion
```

schema/2.0/model/cyclonedx-cryptography-2.0.schema.json
```diff
@@ -325,6 +325,46 @@
...
+                {
+                  "type": "integer",
+                  "minimum": 0
+                },
+                {
+                  "type": "object",
+                  "anyOf": [
+                    {
+                      "required": [
+                        "min"
```
jkowalleck:
```suggestion
```

schema/2.0/model/cyclonedx-cryptography-2.0.schema.json
```diff
@@ -325,6 +325,46 @@
...
+              "oneOf": [
+                {
+                  "type": "integer",
+                  "minimum": 0
+                },
+                {
+                  "type": "object",
+                  "anyOf": [
+                    {
+                      "required": [
```
jkowalleck:
```suggestion
                      "required": [ "min" ], "properties": { "min": true  }
```

schema/2.0/model/cyclonedx-cryptography-2.0.schema.json
```diff
@@ -325,6 +325,46 @@
...
+                        "min"
+                      ]
+                    },
+                    {
+                      "required": [
+                        "max"
+                      ]
```
jkowalleck:
```suggestion
```

schema/2.0/model/cyclonedx-cryptography-2.0.schema.json
```diff
@@ -325,6 +325,46 @@
...
+                  "type": "object",
+                  "anyOf": [
+                    {
+                      "required": [
+                        "min"
+                      ]
+                    },
+                    {
+                      "required": [
+                        "max"
```
jkowalleck:
```suggestion
```

schema/2.0/model/cyclonedx-cryptography-2.0.schema.json
```diff
@@ -325,6 +325,46 @@
...
+                {
+                  "type": "object",
+                  "anyOf": [
+                    {
+                      "required": [
+                        "min"
+                      ]
+                    },
+                    {
+                      "required": [
```
jkowalleck:
```suggestion
                      "required": [ "max" ], "properties": { "max": true }
```

jkowalleck (changes requested):

schema/2.0/model/cyclonedx-cryptography-2.0.schema.json
```diff
@@ -325,6 +325,42 @@
...
+              "oneOf": [
+                {
+                  "type": "integer",
+                  "minimum": 0
+                },
+                {
+                  "type": "object",
+                  "anyOf": [
+                    {
+                      "required": [ "min" ], "properties": { "min": true  }
```
jkowalleck:
```suggestion
                      "required": [ "min" ], "properties": { "min": true }
```

jkowalleck (changes requested):

jkowalleck (approved):
all relevant tests are passing - especially the newlye added ones:

```text
 test /home/runner/work/specification/specification/tools/src/test/resources/2.0/invalid-cryptography-outlen-2.0.json ...
OK.


test /home/runner/work/specification/specification/tools/src/test/resources/2.0/valid-cryptography-outlen-2.0.json ...
OK.

```

PR #960, opened 2026-06-25
Add sample PQC readiness perspective for CBOM

bhess:
Adds a sample perspective demonstrating how the perspective model can apply in a CBOM context, using PQC readiness as the example. It lives in a new perspectives/ folder, as discussed in the last Cryptography WG meeting.

The mappings also exercise a proposed expectation field (values are `present` or `absent`) on perspective mappings that turns a descriptive mapping into a machine-checkable assertion. That field is not yet in the 2.0 schema, so the sample is illustrative pending that discussion.


Mehrn0ush:
Hi,
I was reading this PR against #947.

Since `changeMechanism` is now an array, would “includes runtime-config” fit the rationale better than “runtime-config or restart-required”? Multiple values now mean required steps rather than alternatives.

And if #947 lands first, I assume the path should move from `cryptoProperties.agility` to `component.agility` — is that right?

The inventory/algorithm/agility split makes sense to me. One thing I wondered: should primitive: "unknown" / "other" be treated like missing primitive for the inventory check? And whether protocol cipher suites or cert->signing-algorithm linkage are in scope for a later revision.

I also couldn't find `perspective-expectation-rationale.md` in this PR — is it planned elsewhere, or should the reference in the description come out for now?

bhess:
Thanks for the good feedback @Mehrn0ush .

> Since `changeMechanism` is now an array, would “includes runtime-config” fit the rationale better than “runtime-config or restart-required”? Multiple values now mean required steps rather than alternatives.

Changed the wording accordingly.

> And if #947 lands first, I assume the path should move from `cryptoProperties.agility` to `component.agility` — is that right?

I agree, Changed the path accordingly to component.agility.

> The inventory/algorithm/agility split makes sense to me. One thing I wondered: should primitive: "unknown" / "other" be treated like missing primitive for the inventory check? And whether protocol cipher suites or cert->signing-algorithm linkage are in scope for a later revision.

I agree, modified the rule to exclude "unknown". I think cert algorithms could be valuable as well, need to brainstorm a bit how to implement this as a JSON path expression since this needs bom-ref matching.

> I also couldn't find `perspective-expectation-rationale.md` in this PR — is it planned elsewhere, or should the reference in the description come out for now?

Right, the reference is removed now.

taleodor:
@bhess I'd appreciate if you can take a look at #1067 - if this aligns with your vision here.

PR #990, opened 2026-07-29
Clean check-in of AI/ML. Supersedes #948. 

stevespringett:
This PR supersedes #948, which introduced a number of unnecessary changed files and originated outside of the specification repo. 

- closes #948


stevespringett:
There are multiple things that are currently broken with #948 that this PR will attempt to resolve. This includes references to objects that no longer exist in 2.0, improper use of objects in 2.0, and lack of integration with the component model. This PR will also introduce the required test fixtures (valid / invalid) necessary for the RFC period.

JunHwan-Kwon:
Thanks for moving the AI/ML work into a clean in-repository branch.

The tensor-level affine quantization gap I raised on #948 still seems to apply here. `modelParameter.format` is `additionalProperties: false` and defines `dataType`, `encoding`, and `mimeType`, with no typed representation for scale values, zero-points, or quantization axes. `modelParameter.properties` can hold those as strings, but the schema cannot validate a per-axis scale vector or check its cardinality against the corresponding tensor dimension.

The reproducible TFLite evidence, the ONNX correspondence (per-tensor, per-axis, and blocked granularity), and a proposed framework-neutral shape are all in that comment:
https://github.com/CycloneDX/specification/pull/948#issuecomment-5116167012

I understand that the draft is still being stabilized, so no response is needed at this stage. Once the structure and fixtures settle, I would be glad to help with the aligned JSON, XML, and protobuf definitions and the corresponding valid and invalid tests.

JunHwan-Kwon:
@stevespringett @mrutkows One small integration check following my review on CycloneDX/cyclonedx-property-taxonomy#175.

After #990 was opened, commit `mrutkows/specification@e1116c4` on #948 added a `properties` array inside `modelParameter.format`, and the affine example there used that extension point. In the current branch behind #990, `format` remains `additionalProperties: false` with `dataType`, `encoding`, and `mimeType`, while the sibling `modelParameter.properties` extension point is present.

Was the later `format.properties` addition intended to be carried forward into #990, or is the sibling `modelParameter.properties` array now the intended parameter-level extension point?

Either location would preserve named-parameter binding. I am asking only so that the quantization worked examples in CycloneDX/cyclonedx-property-taxonomy#175 use the intended canonical location.

stevespringett:
@mrutkows can you please take a look and see if this PR matches your expectation of the way the new AI model should look. Also look at the test cases.

NOTE: This PR makes assumptions that other PRs (currently in the public RFC period and not yet approved by TC54) have received positive votes and have been merged.

mrutkows:
@stevespringett Perhaps we should meet to review to assure this gets the dedicated time needed for closure.

Here are the summary of changes that need review:

1 · Core Concept Change
modelCard is renamed to modelProperties. The description makes the intent explicit: intrinsic/technical characteristics stay here; contextual, ethical, and fairness concerns are pushed to a separate risk model; audience-tailored views (model cards) are expressed via perspectives. This is a clean separation-of-concerns aligned with v2.0 architecture.

Matt: modelCard is an industry term and IMO we would regret not adopting the term as a grouping (despite alternate "views" being possible via CDX) and taking a stand on the elements that make up a modelCard.

2 · Architecture Definitions Refactored

The nested structural { primary, secondary, topologyType } object inside modelArchitecture is replaced by three standalone top-level definitions: architectureFamily, architectureFeature, and architectureTopology. Each uses a oneOf (predefined enum | custom extension object), which is cleaner than the old anyOf with regex patterns.

Matt: Nesting of the 3 elements of a model arch. seemed cleaner...

3 · Risk/Ethics Definitions Removed
riskGroup, consideration, fairnessAssessments, ethicalConsiderations, designConsiderations are all gone — deferred to cyclonedx-risk-2.0.schema.json. ⚠️ That file is currently untracked.

Matt: Assume these are moved elsewhere?

4 · modelParameter Massively Enriched
The old parameter had only name, format, description. Steve adds: modality, dataType, mimeType, quantization, shape, processingStage, required, defaultValue, and constraints.

Matt: I believe that aligned with the additions proposed in the other PR...  have to verify 1-by-1

5 · Measure Types Moved to Common Schema
timeUnits, energyUnits, co2Units, measure, energyMeasure, co2Measure are all removed from this file and now referenced from cyclonedx-common-2.0.schema.json.

Matt: assume these was moved elsewhere as you conveyed the last few months...

6 · Bug Fixes
Multiple broken $ref paths corrected (missing $ in #/defs/, wrong path separators, #/definitions/ → #/$defs/).

TYVM

Key Discussion Points:
1. behavioralParadigm and specializedProcessing removed with no replacement

Matt: Discuss if this really disappeared?

3. performanceMetric.type changed from a 23-value enum to an open string — less validation enforcement

Matt: Discuss? enum. may be too untenable, but CDX property perhaps?

5. Custom cdx: unit extension pattern removed from performanceMeasure

Matt: Was really hoping to go in this direction to assure we limit property values to those property names that match (to encourage correctness of information); otherwise any URN value works and prevents at some validation of correctness.  Understand that this is a new precedent...

7. Dependency on common schema definitions that need to be verified as present.

Matt: I have no idea how to contain this other than a full review once we have most of 2.0 schema finalized...

mrutkows:
Another "Advisor" reiterated the points of discussion:

1. Confirm the risk model / perspectives concepts exist (or are landing) in core v2.0, and whether they actually cover groups-at-risk, harms/benefits, and fairness assessments at the granularity your consideration object had — **that's the one change with real content-loss risk if those core objects turn out thinner than what you removed.**

2. Confirm modelProperties (no bom-ref) vs. your modelCard (had a bom-ref) is intentional given the new xxxProperties convention — if some other part of the spec still expects an addressable model card, that's a breaking change worth surfacing now.

3. Whether dropping the curated benchmark/metric enum is a considered tradeoff (open strings avoid taxonomy drift) or an oversight from schema simplification.

mrutkows (changes requested):
Would like to discuss the comments I added in the PR's conversation...

schema/2.0/model/cyclonedx-ai-ml-2.0.schema.json
```diff
@@ -0,0 +1,1354 @@
...
+        },
+        "properties": {
+          "title": "Properties",
+          "description": "Provides the ability to document name-value pairs for the model. Property names of interest to the general public are encouraged to be registered in the [CycloneDX Property Taxonomy](https://github.com/CycloneDX/cyclonedx-property-taxonomy). Formal registration is optional.",
+          "$ref": "cyclonedx-common-2.0.schema.json#/$defs/properties"
+        }
+      }
+    },
+    "learningType": {
+      "title": "Learning Type",
```
HesAnEasyCoder, line 112: Why is this repeated when already listed above?  Why is this repeated when already listed above?  Maybe take away the "Learning Type" above?

schema/2.0/model/cyclonedx-ai-ml-2.0.schema.json
```diff
@@ -0,0 +1,1354 @@
...
+              "type": "string",
+              "title": "Description",
+              "description": "A description of the custom learning paradigm."
+            }
+          }
+        }
+      ]
+    },
+    "modelTask": {
+      "title": "Model Task",
```
HesAnEasyCoder, line 156: Why is this repeated when already listed above?  Maybe take away the "task" above?

stevespringett:
This branch has been synced with 2.0-dev now that many other development branches have been approved and merged into it.

taleodor:
I share Matt's concern about losing "Model Card" as the industry-adopted term. That's usually the first thing that comes up in relevant conversations.

Question / suggestion - would it be possible to add optional type parameter (enum) to perspective, where initial value could be "model-card". And additionally, can we bake such model-card perspective into the spec? Somewhat similar to #960, but with the addition of type.

P.s. AI found a reference to modelCard in perspective schema here - https://github.com/CycloneDX/specification/blob/c257f85a1682205c02c4e65d375c2ae0898fcad2/schema/2.0/model/cyclonedx-perspective-2.0.schema.json#L84  Probably should be updated.

jkowalleck:
there are some conflicts that require fixing.

schema/2.0/model/cyclonedx-ai-ml-2.0.schema.json
stevespringett: The title is for the human-readable documentation that will be submitted to ECMA and ISO.

schema/2.0/model/cyclonedx-ai-ml-2.0.schema.json
stevespringett: The title is for the human-readable documentation that will be submitted to ECMA and ISO.

JunHwan-Kwon:
I followed up on the recent WG discussion of quantized model variants and finer-grained quantization records by measuring two revision-pinned Granite 4.2 3B artifacts carrying the same `Q4_K_M` release label: IBM's publication and Bartowski's publication.

Both expose the same 363 tensor names with matching shapes and 3,659,737,600 serialized tensor elements each, but 100 tensor-to-encoding assignments differ. The IBM artifact contains Q4_K/Q6_K/F32 = 241/41/81 tensors; the Bartowski artifact contains Q4_K/Q5_K/Q6_K/F32 = 141/80/61/81.

The primary comparison uses the upstream `gguf-py` reader from the `gguf` 0.19.0 Python package. It records a serialized-descriptor projection hash separately from an order-independent name-to-encoding assignment hash. A second implementation I maintain was used only as a cross-check and reproduced the descriptor hashes and normalized counts; the comparison above does not depend on it.

At #990 head `38dfe9c`, `modelProperties.quantization` is one object for the distributed weights, while parameter-level quantization is attached to model inputs and outputs. These measurements leave one representation-boundary question: should heterogeneous stored-weight assignments be directly queryable in a BOM, referenced through a content-addressed external inventory, or intentionally remain outside this schema surface?

If this level of detail belongs in or alongside the BOM, my initial preference would be a compact in-BOM summary linked by hash to the exact external assignment inventory, rather than requiring full per-tensor listings inline. I would appreciate guidance on whether that fits the intended boundary.

I am not suggesting that a parser-specific tensor dump be copied into the specification. The evidence package separates serialized observations from declared recipe and lineage information, and does not infer the conversion command, importance-matrix consumption, calibration provenance, or model quality from artifact structure.

Reproduction evidence: [artifact identities and hashes](https://github.com/JunHwan-Kwon/tensor-quantization-metadata-study/blob/evidence-pr990-38dfe9c-gguf-v1/experiments/gguf-imatrix/artifacts.json), [measurement protocol](https://github.com/JunHwan-Kwon/tensor-quantization-metadata-study/blob/evidence-pr990-38dfe9c-gguf-v1/experiments/gguf-imatrix/protocol.md), [exact assignment comparison](https://github.com/JunHwan-Kwon/tensor-quantization-metadata-study/blob/evidence-pr990-38dfe9c-gguf-v1/experiments/gguf-imatrix/results/tensor-assignment-comparison.json), [secondary-parser cross-check](https://github.com/JunHwan-Kwon/tensor-quantization-metadata-study/blob/evidence-pr990-38dfe9c-gguf-v1/experiments/gguf-imatrix/crosschecks/deepbom/normalized-comparison.json), and [draft-schema observations](https://github.com/JunHwan-Kwon/tensor-quantization-metadata-study/blob/evidence-pr990-38dfe9c-gguf-v1/experiments/gguf-imatrix/standards-mapping/pr-990/schema-observation.json).

mrutkows:
> should heterogeneous stored-weight assignments be directly queryable in a BOM, referenced through a content-addressed external inventory, or intentionally remain outside this schema surface?

Yes, I believe per-layer (and per-module) weights should be expressible and what I have been implying since not being able to address this in the 2.0 schema as we just simply did not have the time to develop that level of detail in the time given.  It is my hope that it is something we can and should discuss and develop for the 2.1 time frame and attempt to bring in more model experts to have the proper discussions and examples to design it correctly.

JunHwan-Kwon:
@mrutkows Thank you for clarifying the scope for 2.0. I am happy to carry the per-layer and per-module weight examples into the proposed 2.1 discussion, rather than request an expansion of #990 at this stage.

I also applied the same tensor-assignment comparison procedure to a second revision-pinned pair. Within each pair, both publications carry the `Q4_K_M` release label. The Granite pair is divergent: 100 of 363 tensor-to-encoding assignments differ. The Qwen3-4B pair is concordant: all 398 tensor names, shapes, and per-tensor encodings match, and the name-to-encoding assignment hashes are identical.

The Granite result shows that matching release labels do not establish assignment equality; the Qwen result provides a concordant comparison under the same procedure. These observations concern encoding assignments, not numerical weight equivalence or identical artifact bytes, and do not estimate how often either outcome occurs.

I will retain the exact inventories, hash definitions, and examples for the follow-up discussion and consider this stored-weight assignment question resolved for the 2.0 scope.

Reproduction evidence: [pinned artifacts](https://github.com/JunHwan-Kwon/tensor-quantization-metadata-study/blob/evidence-pr990-38dfe9c-gguf-v3/experiments/gguf-paired-replication/qwen3-4b/artifacts.json), [protocol](https://github.com/JunHwan-Kwon/tensor-quantization-metadata-study/blob/evidence-pr990-38dfe9c-gguf-v3/experiments/gguf-paired-replication/qwen3-4b/protocol.md), and [assignment comparison](https://github.com/JunHwan-Kwon/tensor-quantization-metadata-study/blob/evidence-pr990-38dfe9c-gguf-v3/experiments/gguf-paired-replication/qwen3-4b/results/tensor-assignment-comparison.json).

jkowalleck:
@stevespringett could you fix these merge conflicts, before we do a review on this?

jkowalleck:
RFC notice sent on September 16, 2026.
Public RFC period ends October 14, 2026.

schema/2.0/model/cyclonedx-ai-ml-2.0.schema.json
```diff
@@ -0,0 +1,1030 @@
...
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://cyclonedx.org/schema/2.0/model/cyclonedx-ai-ml-2.0.schema.json",
+  "type": "null",
+  "title": "CycloneDX AI/ML Model",
+  "$comment": "OWASP CycloneDX is an Ecma International standard (ECMA-424) developed in collaboration between the OWASP Foundation and Ecma Technical Committee 54 (TC54). The standard is published under a royalty-free patent policy. This JSON schema is the reference implementation and is licensed under the Apache License 2.0.",
+  "$defs": {
+    "modelProperties": {
+      "type": "object",
+      "title": "AI/ML Model Properties",
+      "description": "Intrinsic technical characteristics of a machine learning model. This object SHOULD be specified for any component of type `machine-learning-model` and shall not be specified for other component types. Contextual and subjective aspects of a model, including ethical and fairness considerations, potential harms, groups at risk, and impact assessments, are expressed with the risk model and reference the model component. Views of this data tailored to a specific audience, such as a model card, are expressed with perspectives.",
```
mrutkows, line 11: I am a bit unsettled that the work to refactor considerations around risk schema (and developing a "consideration" object to describe the consideration (and harms/benefits related to them) has been completely removed.  I have no idea how to relate or model these relative to the model component itself.  These are, IMO. not "perspective" addressable items.  Again, these are design time elements that need direct relational ties to the actual model component (as well as other considerations like environmental which I have yet to see how the came out the other side).

---

Claude Confirmed: ethicalConsiderations and fairnessAssessments are removed from the v2.0 modular schemas
Yes, you are correct. Both concepts have been deliberately removed from the canonical modular schema files (schema/2.0/model/). The evidence:

ethicalConsiderations — zero hits in the modular schemas. Only survives as a string mention in a description field inside cyclonedx-common-2.0.schema.json (a prose description of what a model card is) and in cyclonedx-ai-ml-2.0.schema.json as explanatory text — not as a schema property.

fairnessAssessments — zero hits in the modular schemas. Completely gone as a structured object.

mrutkows (changes requested):

schema/2.0/model/cyclonedx-ai-ml-2.0.schema.json
stevespringett: Thanks Matt. You are right that the `consideration` object and the `modelCard.designConsiderations` nesting are gone. The information they carried is not. It moved into the risk model, which is a root-level container. A perspective does not hold any of this data. The model card perspective only selects it.

**How a consideration ties to the model.** `risk.affects` references the model component by `bom-ref`, and also references the persona party that represents the group at risk. That is a direct, typed relationship to the model, and it is the same direction the vulnerability model has always used: a vulnerability names the components it affects, and a component does not list its vulnerabilities. From `valid-ai-ml-risk-integration-2.0.json`:

```json
{
  "bom-ref": "risk-demographic-disparity",
  "name": "Demographic performance disparity",
  "domains": [ { "type": "ethical" }, { "type": "safety" } ],
  "affects": [ "model-1", "party-elderly-users" ],
  "relatedWeaknesses": [ "weakness-age-representation" ],
  "inherentRisk": {
    "likelihood": { "level": "high" },
    "impacts": [
      { "level": "major", "polarity": "harm", "categories": [ "fairness", "health" ] },
      { "level": "moderate", "polarity": "benefit", "categories": [ "health" ],
        "description": "Where summaries are accurate, clinicians review notes for this group faster, shortening time to treatment." }
    ]
  },
  "externalReferences": [
    { "type": "evidence", "url": "https://acme.example.com/responsible-ai/audits/demographic-disparity-2026q3" }
  ]
}
```

Referencing rather than nesting was deliberate. A single consideration often concerns several subjects at once: the model, the training dataset, the deployment, and one or more affected groups. Nested under one model card, it would have to be copied into each of them. An assessment also commonly covers several models at once.

**Where each #948 field went:**

| #948 | #990 |
|---|---|
| `consideration.name`, `description` | `risk.name`, `statement`, `description` |
| `groupsAtRisk` (`riskGroup`) | `risk.affects` referencing a persona party (a party also carries `externalReferences`) |
| `harms`, `benefits` | `rating.impacts[]`, each with `polarity` harm or benefit, `categories`, and `description`; one risk can carry both |
| nested `risk`, `assessment` | first-class risks and `risks.assessments`, referenced by `bom-ref` |
| `ethicalConsiderations` | risks with domain `ethical` |
| `fairnessAssessments` | assessments of type `ai-impact` (or a custom type) over risks with domain `fairness`, plus evaluation metrics with a `slice` in `component.measurements` |
| `environmentalConsiderations` | `component.measurements[]` of `energyConsumption`: every #948 field plus `description`, `formula`, `graphics`, and `externalReferences`; environmental harm itself is a risk with domain or impact category `environmental` |
| `technicalLimitations` | `modelProperties.limitations`; a limitation that exposes a group to harm is also recorded as a risk |
| `users`, `useCases` | `component.parties` with role `end-user`, and `component.useCaseAssertions` |


**What #990 does not carry over**:

1. `performanceTradeoffs` has no dedicated slot. A trade-off is a risk with a harm impact and a benefit impact, or a limitation.
2. `technicalLimitations` entries are strings, not objects with groups, harms, and benefits. Anything with those parts is a risk.

Your comment surfaced two gaps, both now fixed on the branch:

1. A #948 consideration could carry `externalReferences`, but a `risk` could not; only an assessment could. `risk` now has `externalReferences`, so evidence such as an audit report can be cited on each consideration. The fixture cites one.
2. The fixture's "Ethical Considerations" mapping selected every ethical risk in the document, not only the risks that affect this model. It is now `$.risks.risks[?(@.affects[?(@=='model-1')] && @.domains[?(@.type=='ethical')])]`, so the model card follows the `affects` link. The `modelProperties` description also names `affects` as the link now, rather than saying only "the risk model".

glenmessenger:
We operate k8s-aibom, an open-source Kubernetes controller that emits CycloneDX 1.6 ML-BOMs for AI workloads at runtime, and since v1.5.0 it cryptographically verifies model signature claims (Sigstore/OMS bundles verified against configurable trust roots, with Rekor transparency-log inclusion). Reading this PR's evidence model against that implementation, we hit one gap we couldn't express:

There is no analysis technique for cryptographic signature verification as identity evidence. The existing `signature` technique is defined as byte/code pattern matching and is excluded from identificationMethod; `attestation` is a party assertion rather than direct analysis; `hash-comparison` is a bare digest check. But "this model's identity was established by verifying a signature chain against an operator-configured trust root, with transparency-log inclusion" is direct analysis, is an identification method, and in practice is the *strongest* identity evidence an ML-BOM can carry — model files otherwise carry no intrinsic identity (no manifest, no package coordinates), which makes this vocabulary gap more consequential for AI/ML components than for software packages.

Proposal: add a `signature-verification` technique (meta:enum along the lines of "Cryptographic verification of a signature over the subject against a trust root, including any transparency-log inclusion check, that binds an asserted identity to the subject's content") and permit it in identificationMethod. We'd suggest its `value` carry the verified signer identity, since that's what a downstream consumer needs to act on.

Happy to contribute worked examples from production output if useful — we currently express these facts as custom properties for exactly this reason, and would much rather emit them in the specification's own vocabulary.

glenmessenger:
A second, smaller question from the same implementation experience: the new AI/ML schema describes the model thoroughly (architecture, quantization, training, metrics) but has no integrity or identity surface of its own — presumably by design, with model-as-component composing with componentEvidence/componentIdentityEvidence for identity, confidence, and occurrences. If that's the intent, could the model README or an example make the composition explicit (a model component whose identity carries evidence with per-technique confidence)? In our experience producing runtime ML-BOMs, identity confidence is the field auditors interrogate first — a canonical example would keep the ecosystem's ML tooling from treating the AI/ML schema as self-contained and skipping evidence entirely.

Relatedly, is the intent that as-deployed vs. as-trained context be carried by the perspectives work (#1067)? Runtime-generated ML-BOMs need somewhere to say "this describes what is serving, not what was built," and we'd rather align with the intended mechanism than invent one.

PR #1000, opened 2026-08-03
2.0 tests init

jkowalleck:
- [x] Java tests - functional - validate test resources with schema
- [x] JS tests - functional - validate test resources with schema
  - found some false-positives/false-negatives - due to errors in test resources
- [x] JS tests - schema validate
  - includes https://github.com/CycloneDX/specification/issues/996
  - includes https://github.com/CycloneDX/specification/issues/992
  - `type=object` set explicitely 
- JS tests - semantics
  - test: no self-$ref by file
  - test: refType usage (`bom-ref` <-> refType) - https://github.com/CycloneDX/specification/issues/998
  - additionalProperties is `false` - https://github.com/CycloneDX/specification/issues/995
- [x] PHP tests - functional - validate test resources with schema

added tests only,
no schema/res fixes - they are prepared in 
- https://github.com/CycloneDX/specification/pull/997 
- https://github.com/CycloneDX/specification/pull/1001

PR #888, opened 2026-03-16
streamline self refs

jkowalleck:

PR #1001, opened 2026-08-04
[2.0] fixes on schema and test resources

### LINKED ISSUE #992: [2.0]: fix unreachable cross-references
in CycloneDX 2.0 we have unreachable cross-references

- https://github.com/CycloneDX/specification/blob/2a860677bc58cc9accd7c8810eee1b4f29ce3343/schema/2.0/model/cyclonedx-license-2.0.schema.json#L128
- https://github.com/CycloneDX/specification/blob/2a860677bc58cc9accd7c8810eee1b4f29ce3343/schema/2.0/model/cyclonedx-cryptography-2.0.schema.json#L97
- https://github.com/CycloneDX/specification/pull/678/changes#r3685780638
- to be continued 

## main goal

have these links fixed

## additional goal

have a tool that checks all those `$ref` values 
- minimum: that start with a `.` and see if they exist - not only the file, but also the linked element
- maximum: any value and see if the linked element exists - in the remote document (starts with `./` or `../`) or the same document (start with `#`) ... 

hm, Our validators that test weather the test cases are schema-valid should have picked this up already ...
let's see why they did not

### LINKED ISSUE #995: [2.0] assert every object has `additionalProperties=false`
in CycloneDX 2.0 we have a lot of new `type=object`.

## goal

assert that every `object` as `additionalProperties=false` set

## tooling

have a QA tool that asserts this `additionalProperties `is present and set to `false`

### LINKED ISSUE #996: [2.0] assert every object's `required` properties actually exist
in CycloneDX 2.0 we have huge amount ob new object structures built.

## goal 

test that every `required` property exists.

- defined in `.required`
- defined in `{if,then,else}...required`
- any other reference to a property that must exist for the schema to be sane

## tooling

have a QA tool that asserts this `additionalProperties `is present and set to `false`

### LINKED ISSUE #998: [2.0] add semantictests: all `bom-ref` must be of type `refType`
## goals: 
- add a test, that checks that ALL properties `bom-ref` properties are of type `cyclonedx-common-2.0.schema.json#$defs/refType`
- add a test, that checks that ONLY properties `bom-ref` properties are of type `cyclonedx-common-2.0.schema.json#$defs/refType`

## automation: 
- tests must run on every PR
- tests must be executable locally 

## out of scope: 
- altering schemas or tests - done in different issue/PR

jkowalleck:
fixes or related to 

- https://github.com/CycloneDX/specification/issues/992
- https://github.com/CycloneDX/specification/issues/995
- https://github.com/CycloneDX/specification/issues/996
- https://github.com/CycloneDX/specification/issues/998

see also:
-  https://github.com/CycloneDX/specification/pull/888

besides those fixes above, this is in the PR:
- made the `refLinkType` inherit explicit
- fixes the array nesting of `extensibleProerties`
- consolidated milisecond timestamp for `JSS`
- adds `$comments`
- enhanced some docs 
- added test cases
  - for `extensibleProperties` 
- fixed valid test data 
  - all test data that was just carried over from CDX 1.x was transferred to CDX 2.0schema  
- fixed invalid test data
  - removed `invalid-component-choice-missing-name-2.0.json` - as the `name` was optional or did not exist.  
    :question: was this intentional/correct

tests for the things are based on: 
- https://github.com/CycloneDX/specification/pull/1000


tools/src/test/resources/2.0/invalid-extensible-properties-item-wrong-prefix.json
jkowalleck, line 1: added new test file

tools/src/test/resources/2.0/valid-extensible-properties.json
jkowalleck, line 1: added new test file

schema/2.0/model/cyclonedx-jss_X590_2023_10-2.0.schema.json
```diff
@@ -192,18 +205,16 @@
           "type": "string"
         },
         "created": {
-          "$ref": "#/$defs/timestamp",
+          "$ref": "#/$defs/timestamp/$defs/withMilliseconds",
           "title": "Created Timestamp",
           "description": "The time at which this signature was originally created. The creator can use any time it deems most appropriate as the time the signature was created, but it MUST be precise to the nearest millisecond (exactly three digits after the decimal place in seconds). The created property MUST NOT be changed when creating a new version of the signature.",
-          "$comment": "Normative source: ITU-T X.590 clause 6.3.1",
```
jkowalleck, line 198: consolidated to `#/$defs/timestamp/$defs/withMilliseconds`

schema/2.0/model/cyclonedx-jss_X590_2023_10-2.0.schema.json
```diff
@@ -192,18 +205,16 @@
...
         },
         "modified": {
-          "$ref": "#/$defs/timestamp",
+          "$ref": "#/$defs/timestamp/$defs/withMilliseconds",
           "title": "Modified Timestamp",
           "description": "The time that this particular version of the signature was last modified. The creator can use any time it deems most appropriate as the time that this version of the signature was modified, but it MUST be precise to the nearest millisecond (exactly three digits after the decimal place in seconds). The modified property MUST be later than or equal to the value of the created property. If the created and modified properties are the same, then this is the first version of the signature.",
-          "$comment": "Normative source: ITU-T X.590 clause 6.3.1. Cross-field invariant: 'modified' MUST be greater than or equal to 'created'. JSON Schema 2020-12 cannot compare two sibling property values, so this constraint is NOT enforced by this schema and MUST be validated by the consuming tool or conformance suite.",
-          "pattern": "^[0-9]{4}-(?:0[1-9]|1[0-2])-(?:0[1-9]|[12][0-9]|3[01])T(?:[01][0-9]|2[0-3]):[0-5][0-9]:[0-5][0-9]\\.[0-9]{3}Z$"
```
jkowalleck, line 206: consolidated to `#/$defs/timestamp/$defs/withMilliseconds`

tools/src/test/resources/2.0/invalid-component-choice-missing-name-2.0.json
jkowalleck, line 1: file was removed, since the relevant test case did not exist. 
the `componentCHoice` does not require/have a `name` 

❓ was this intentional/correct

jkowalleck:
had Fable 5 analyze the state of the schema.

my query:

```text
Analyze the JSON Schema structure under https://github.com/CycloneDX/specification/blob/2.0-tests-init-fixes/schema/2.0/cyclonedx-2.0.schema.json . Ignore all `-bundled` files.

Starting from the root schema , follow every $ref transitively through the files in  (e.g. cyclonedx-metadata, cyclonedx-component, cyclonedx-service, cyclonedx-dependency, cyclonedx-composition, cyclonedx-vulnerability, cyclonedx-annotation, cyclonedx-formulation, cyclonedx-declaration, cyclonedx-definition, cyclonedx-citation, cyclonedx-perspective, cyclonedx-common, cyclonedx-party, cyclonedx-license, cyclonedx-data, cyclonedx-cryptography, cyclonedx-ai-modelcard, cyclonedx-release-notes, cyclonedx-standard, cyclonedx-patent, cyclonedx-jss_X590_2023_10, cyclonedx-api) and check whether the schema makes sense in its current structure. Specifically verify:
Code

- Broken references: every `$ref` (relative file path + JSON pointer like `#/$defs/xyz`) resolves to an existing file and an existing $defs entry in that file. List any $ref that points to a missing file or missing definition.
- `$id` consistency: each file's $id matches its actual path/filename and the expected https://cyclonedx.org/schema/2.0/... convention.
Orphaned files/definitions: model files or $defs entries not reachable from the root schema (excluding the separate cyclonedx-api-2.0.schema.json entry point — check its refs too).
- Circular references that could be problematic, and duplicate definitions of the same concept in multiple files.
- Draft 2020-12 correctness: proper use of $defs, unevaluatedProperties interaction with allOf/$ref (note the root uses unevaluatedProperties: false together with allOf extensibleProperties — check whether this composition works as intended across files), missing $schema declarations, wrong keyword usage, invalid regex patterns, enum/const misuse.
- Naming/structural inconsistencies: inconsistent def naming, inconsistent title/description conventions, anything that looks off (e.g. typos in filenames like jss_X590 vs X.509).

Produce a report of concrete issues found, each with file path, JSON pointer location, and a suggested fix. Also state explicitly which aspects were checked and found to be fine.

I need the report in pastable markdown.

```

this are the results: 
----
# CycloneDX 2.0 JSON Schema Structure Review

**Scope:** branch `2.0-tests-init-fixes`, root `schema/2.0/cyclonedx-2.0.schema.json` plus all 22
`schema/2.0/model/*.schema.json` files and `schema/2.0/cyclonedx-api-2.0.schema.json`.
All `*-bundled*` files were ignored. All 433 `$ref` occurrences were resolved transitively;
every file was validated against the JSON Schema 2020-12 meta-schema.

---

## 1. Concrete issues found

### 1.1 References that escape the `schema/2.0/` tree (and have `$id` mismatches)

| # | File | Location (JSON pointer) | Issue |
|---|------|-------------------------|-------|
| 1 | `schema/2.0/model/cyclonedx-cryptography-2.0.schema.json` | `#/$defs/cryptoProperties/properties/algorithmProperties/properties/algorithmFamily/$ref` and `.../ellipticCurve/$ref` | `"../../cryptography-defs.schema.json#/definitions/…"` points **outside the versioned 2.0 tree** to `schema/cryptography-defs.schema.json`. The pointer itself resolves (the target has `definitions/algorithmFamiliesEnum` and `definitions/ellipticCurvesEnum`), **but**: (a) the target is a **draft-07** schema while all 2.0 schemas are 2020-12; (b) resolving the relative ref against the model file's `$id` yields `https://cyclonedx.org/schema/cryptography-defs.schema.json`, while the target declares `$id: http://cyclonedx.org/schema/cryptography-defs.schema.json` (**http vs https mismatch** — strict resolvers that key their registry by `$id` will fail); (c) it is unversioned, so 2.0 validation results can change whenever the shared file is regenerated. **Fix:** copy/generate a versioned, 2020-12 `$defs`-based snapshot into `schema/2.0/` (e.g. `schema/2.0/cyclonedx-cryptography-defs-2.0.schema.json`) with an `https://cyclonedx.org/schema/2.0/...` `$id`, and reference that. |
| 2 | `schema/2.0/model/cyclonedx-license-2.0.schema.json` | `#/$defs/license/properties/id/$ref` | `"../../spdx.schema.json"` — same class of problem: target is draft-07, unversioned, outside `schema/2.0/`, and declares `$id: http://cyclonedx.org/schema/spdx.schema.json` while the ref resolves to the `https://` URI. **Fix:** ship a versioned SPDX enum snapshot under `schema/2.0/` (as 1.x does with a sibling `spdx.schema.json` in the same directory) or at minimum align the `$id` scheme. |

These two are the only refs that do not stay within `schema/2.0/`; **every other `$ref` (relative path + `#/$defs/...` pointer) resolves to an existing file and an existing definition.**

### 1.2 Orphaned / dead definitions

| # | File | Location | Issue & suggested fix |
|---|------|----------|-----------------------|
| 3 | `model/cyclonedx-common-2.0.schema.json` | `#/$defs/organizationalEntityOrContact` | Content is literally `{ "$comment": "TODO" }` — an unfinished stub, referenced by nothing. As a schema it is `true` (accepts anything). **Fix:** implement it (e.g. `anyOf` of `organizationalEntity` / `organizationalContact`) or delete it before release. |
| 4 | `model/cyclonedx-common-2.0.schema.json` | `#/$defs/baseObject` | A "base object for all CycloneDX entities" mixin (extensibleProperties + `properties` + `externalReferences`) that **no schema composes**. Its own `$comment` describes an architecture ("concrete leaf schemas composing this mixin must close themselves with `unevaluatedProperties: false`") that is not implemented anywhere. **Fix:** either wire it into the entity definitions or remove it. See also issue 8 — this orphan is the symptom of a larger design gap. |
| 5 | `model/cyclonedx-common-2.0.schema.json` | `#/$defs/base64` | Unreferenced. `attachment.content` is a plain `string` and does not use it even when `encoding: "base64"` is set. **Fix:** reference it from `attachment` (e.g. via `if/then` on `encoding`) or remove it. |
| 6 | `model/cyclonedx-common-2.0.schema.json` | `#/$defs/timestamp` | Unreferenced. Every consumer needing a timestamp uses inline `format: date-time`, and the JSS file carries its own, stricter `timestamp` def. **Fix:** make the model files reference `common#/$defs/timestamp` (or delete it). Note its regex is also weaker than the JSS one (accepts `2023-13-99T…`) — see issue 10. |
| 7 | `schema/2.0/cyclonedx-api-2.0.schema.json` | whole file | An empty shell: `type: object`, empty `$defs`, no `properties`/`required` — it validates **any** JSON object. The README describes it as "the normative API-focused schema", which it currently isn't. It references no model files at all. **Fix:** flesh it out or clearly mark it as a placeholder in the README. |

There are **no orphaned model files** — all 22 model files are reachable from the root schema.

### 1.3 `unevaluatedProperties` / extensibility composition

| # | File | Location | Issue & suggested fix |
|---|------|----------|-----------------------|
| 8 | root + all model files | root `#/allOf` + `#/unevaluatedProperties`; e.g. `component`, `service`, `vulnerability`… defs | The root composition itself is **correct**: in draft 2020-12, `unevaluatedProperties: false` *does* see through `allOf`/`$ref`, so root-level `properties` **and** the `ext:*` `patternProperties` from `extensibleProperties` both remain allowed while everything else is rejected. **However, the pattern is only applied at the document root.** All nested entity objects (component, service, vulnerability, party, hash, etc. — 100+ objects) are closed with `additionalProperties: false` and do **not** compose `extensibleProperties`. Consequently `ext:*` extension properties are legal *only* at the BOM top level, contradicting the `extensibleProperties` description ("enables organizations … to safely introduce custom properties") and the mixin comments in `common` (issues 3–4). **Fix:** decide the intended granularity. If entities should be extensible: give each entity `allOf: [{$ref: …extensibleProperties}]` and replace `additionalProperties: false` with `unevaluatedProperties: false` (this is exactly what the orphaned `baseObject` was built for). If only the root should be extensible: update the `extensibleProperties` description/comments and delete `baseObject`. |
| 9 | `model/cyclonedx-jss_X590_2023_10-2.0.schema.json` | `#/$defs/signatureObject/unevaluatedProperties` | `unevaluatedProperties: true` is a no-op — it is identical to omitting the keyword. The `$comment` explains the openness is intentional (X.590 §6.2 allows extra metadata), so the *behaviour* is fine, but the keyword adds noise and suggests a misunderstanding. **Fix:** drop the keyword and keep the `$comment`, or keep it only if you want an explicit marker (then say so in the comment). |

### 1.4 Duplicate definitions of the same concept

| # | Concept | Files | Issue & suggested fix |
|---|---------|-------|-----------------------|
| 10 | `timestamp` | `common#/$defs/timestamp` vs `jss_X590…#/$defs/timestamp` | Two different regexes for the same concept: common's `^\d{4}-\d{2}-\d{2}T…$` accepts invalid dates (`2023-13-99`); JSS's validates month/day/hour ranges. Common's is also completely unused (issue 6). **Fix:** keep one canonical def (JSS one is normative to X.590 and may stay, but common should adopt equivalent strictness) and reference it. |
| 11 | `postalAddress` | `common#/$defs/postalAddress` vs `party#/$defs/postalAddress` | Two **divergent** address models are both in active use: party's adds `isoCode` and `coordinates`; common's is used from `organizationalEntity`. Same title, different shapes → the same real-world address validates differently depending on which parent object holds it. **Fix:** keep one (party's superset) and have `common` `$ref` it, or move it wholly into `common`. |
| 12 | `identifier` | `component`, `party`, `jss_X590…` (each `#/$defs/identifier`) | Three unrelated concepts share one def name: component = "identifiers asserted by a party", party = "authority-issued identifier (scheme/value)", JSS = "UUID". Not a resolution problem (refs are file-scoped) but confusing for readers and for bundlers/codegen that flatten names. **Fix:** rename to `assertedIdentifier` / `authorityIdentifier` / `signatureId` (or similar). |
| 13 | `hashAlgorithm` | `common` vs `jss_X590…` | common = enum of `SHA-256`-style names; JSS = free-form IANA lowercase (`sha-256`) pattern. Same name, incompatible value spaces. Justifiable (X.590 mandates IANA names) but deserves a rename (e.g. `ianaHashAlgorithm` in JSS) or a `$comment` cross-reference. |
| 14 | `mediaType` | `common#/$defs/mediaType` vs `component#/$defs/component/properties/mime-type` | Component's `mime-type` re-declares the identical pattern `^[-+a-z0-9.]+/[-+a-z0-9.]+$` inline instead of `$ref`-ing `common#/$defs/mediaType`; the property is also named `mime-type` while everything else says "media type". **Fix:** `$ref` the common def; consider renaming the property (breaking) or at least noting the alias. |
| 15 | party vs organizationalEntity | `party-2.0` vs `common-2.0` | Two parallel "who" models coexist: the rich `party`/`organization`/`person`/`persona` model (referenced from exactly one place: `component#/$defs/…parties`) and the legacy `organizationalEntity`/`organizationalContact` (referenced ~20×, incl. `metadata`). If `party` is the 2.0 direction, most of the spec doesn't use it yet. **Fix:** converge (e.g. make `organizationalEntity` an alias/`$ref` into the party model), or document the intended split. |

### 1.5 Naming / structural inconsistencies

| # | File / location | Issue & suggested fix |
|---|-----------------|-----------------------|
| 16 | `model/cyclonedx-jss_X590_2023_10-2.0.schema.json` (filename + `$id`) | The standard is **ITU-T X.590** (JSON Signature Scheme) — the filename drops the dot (`X590`) and mixes `snake_case` (`jss_X590_2023_10`) into an otherwise all-kebab-case directory (`ai-modelcard`, `release-notes`, …). It is *not* a typo for X.509, but it reads like one. **Fix:** rename to e.g. `cyclonedx-jss-x590-2.0.schema.json` (edition date in the file's `$comment`/`description`, not the filename) and update `$id` plus the one inbound ref from `common`. |
| 17 | Titles across model files | Convention is `CycloneDX <Thing> Model` (Annotation, Citation, Common, …) but two deviate: `CycloneDX AI Model Card` and `CycloneDX Model for JSON Signature Scheme (JSS)`. **Fix:** normalize (e.g. `CycloneDX AI Model Card Model` is awkward — pick and document one convention). |
| 18 | root `#/properties/specFormat` | Single-value `"enum": ["CycloneDX"]` — in 2020-12 the idiomatic form is `"const": "CycloneDX"`. Same for `ai-modelcard` `energyMeasure.unit` (`["kWh"]`), `co2Measure.unit` (`["tCO2eq"]`) and `common` `attachment.encoding` (`["base64"]`). **Fix:** use `const` (non-breaking, purely stylistic). |
| 19 | root `#/properties/specVersion` | Only `type: string` with an *example* of `"2.0"` — **any** string validates (`"1.4"`, `"banana"`). Every 1.x schema pins this with an enum. **Fix:** `"const": "2.0"` (or an enum if patch versions are anticipated). |
| 20 | `model/cyclonedx-jss_X590_2023_10-2.0.schema.json` `#/$defs/timestamp/$defs/withMilliseconds` | A `$defs` block nested *inside another definition*, referenced via the unusual pointer `#/$defs/timestamp/$defs/withMilliseconds`. Legal, and the `$comment` marks it internal, but it is the only place this idiom is used and some tooling (doc generators, bundlers) handles nested `$defs` poorly. **Fix (optional):** hoist to a top-level `$defs/timestampWithMilliseconds`. |
| 21 | `model/cyclonedx-definition-2.0.schema.json` | The file exists to hold a single `$defs/definitions` wrapper (standards + patents). Also note the root BOM property is itself called `definitions` — a domain term that collides visually with the JSON-Schema keyword; inherited from 1.6, so acceptable, but the def-named-`definitions`-inside-`$defs` in this file compounds the confusion. **Fix (optional):** fold into `standard`/`patent` files or rename the def. |

### 1.6 Circular references

| # | Cycle | Assessment |
|---|-------|------------|
| 22 | `cyclonedx-component-2.0` ⇄ `cyclonedx-ai-modelcard-2.0` (`component → modelCard/graphicsCollection`; `ai-modelcard → component#/$defs/componentData`) | The only **cross-file** cycle. Harmless for validation (recursion is data-bounded), but it defeats simple topological bundling/codegen. **Fix (optional):** move `componentData` (and friends) into `cyclonedx-data-2.0.schema.json`, which already exists for that purpose — this would break the cycle cleanly. |
| 23 | Self-recursion inside `component` (children/pedigree), `formulation` (workflow steps), JSS (`signatureObject` counter-signature chain), `extensibleProperties/propertyValue` | All intentional and well-formed; no action needed. |

---

## 2. Aspects checked and found to be fine

- **`$ref` resolution:** all 433 refs resolve; every `#/$defs/...` pointer targets an existing definition. No dangling pointers within `schema/2.0/`. The only caveats are the two out-of-tree targets (issues 1–2).
- **`$id` consistency:** all 24 files declare `$id` exactly matching `https://cyclonedx.org/schema/2.0/<path>` for their actual path/filename. No mismatches inside the 2.0 tree.
- **`$schema` declarations:** present in every file, all `https://json-schema.org/draft/2020-12/schema`. No missing or mixed-draft declarations *within* the 2.0 tree.
- **Meta-schema validity:** every file passes validation against the 2020-12 meta-schema.
- **Orphaned files:** none — all model files are reachable from the root. Apart from the four `$defs` entries in `common` (issues 3–6), every definition in every file is referenced at least once.
- **`$defs` usage:** consistently `$defs` (never legacy `definitions`) in all 2.0 files; the "definitions" hits in the root and definition model are domain property/def *names*, not the keyword.
- **Root `unevaluatedProperties: false` + `allOf` composition:** semantically correct at the root — draft 2020-12 annotation flow makes the `ext:*` patternProperties from the referenced `extensibleProperties` count as "evaluated". (The gap is coverage, not correctness — issue 8.)
- **Regex patterns:** all 24 `pattern`/`patternProperties` expressions compile and use only ECMA-262-safe constructs (no lookbehind, no `\p` without flags, etc.). Escaping is correct.
- **Enum/const misuse:** no duplicate enum values, no type-inconsistent enums; only the stylistic single-value enums noted in issue 18.
- **Keyword usage:** no draft-04 leftovers (boolean `exclusiveMinimum`, `id`, etc.); `format` values used (`date`, `date-time`, `idn-email`, `iri-reference`, `uri`, `uuid`) are all standard 2020-12 vocabulary; `if/then/else`, `prefixItems`-free arrays, and `patternProperties` are used correctly.
- **Model-file guard convention:** every model file consistently uses root `"type": "null"` so the file root cannot accidentally be used as a document schema — applied uniformly across all 22 files.
- **Dependency graph:** clean layering otherwise — `common` is the shared base (its only dependency is the JSS file), root fans out to the 13 top-level model refs, and the one problematic cycle is issue 22.

---

## 3. Summary

The modular structure is fundamentally sound: refs resolve, `$id`s are consistent, and the files are
valid 2020-12. The material problems are (a) the two **out-of-tree, cross-draft, `http`/`https`-mismatched
external refs** (crypto-defs, SPDX), (b) the **extensibility architecture being only half-built**
(`baseObject` orphaned, `ext:*` only usable at the root while every nested object is closed with
`additionalProperties: false`), (c) the **empty API entry point**, and (d) several **divergent duplicate
definitions** (`postalAddress`, `timestamp`, party-vs-organizationalEntity) that will produce
inconsistent validation of identical data.

tools/src/test/resources/2.0/invalid-component-choice-missing-name-2.0.json
jkowalleck, line 1: removed the test case, since the `name` was  not required at all ...

jkowalleck:
had Fable 5 analyze the state of the schema.

my query:

```text
Analyze the JSON Schema structure under https://github.com/CycloneDX/specification/blob/2.0-tests-init-fixes/schema/2.0/cyclonedx-2.0.schema.json . Ignore all `-bundled` files.

Starting from the root schema , follow every $ref transitively through the files in  (e.g. cyclonedx-metadata, cyclonedx-component, cyclonedx-service, cyclonedx-dependency, cyclonedx-composition, cyclonedx-vulnerability, cyclonedx-annotation, cyclonedx-formulation, cyclonedx-declaration, cyclonedx-definition, cyclonedx-citation, cyclonedx-perspective, cyclonedx-common, cyclonedx-party, cyclonedx-license, cyclonedx-data, cyclonedx-cryptography, cyclonedx-ai-modelcard, cyclonedx-release-notes, cyclonedx-standard, cyclonedx-patent, cyclonedx-jss_X590_2023_10, cyclonedx-api) and check whether the schema makes sense in its current structure. Specifically verify:
Code

- Broken references: every `$ref` (relative file path + JSON pointer like `#/$defs/xyz`) resolves to an existing file and an existing $defs entry in that file. List any $ref that points to a missing file or missing definition.
- `$id` consistency: each file's $id matches its actual path/filename and the expected https://cyclonedx.org/schema/2.0/... convention.
Orphaned files/definitions: model files or $defs entries not reachable from the root schema (excluding the separate cyclonedx-api-2.0.schema.json entry point — check its refs too).
- Circular references that could be problematic, and duplicate definitions of the same concept in multiple files.
- Draft 2020-12 correctness: proper use of $defs, unevaluatedProperties interaction with allOf/$ref (note the root uses unevaluatedProperties: false together with allOf extensibleProperties — check whether this composition works as intended across files), missing $schema declarations, wrong keyword usage, invalid regex patterns, enum/const misuse.
- Naming/structural inconsistencies: inconsistent def naming, inconsistent title/description conventions, anything that looks off (e.g. typos in filenames like jss_X590 vs X.509).

Produce a report of concrete issues found, each with file path, JSON pointer location, and a suggested fix. Also state explicitly which aspects were checked and found to be fine.

I need the report in pastable markdown.

```

report:


# CycloneDX 2.0 JSON Schema Structure Review

**Scope:** branch `2.0-tests-init-fixes`, `schema/2.0/cyclonedx-2.0.schema.json` + `schema/2.0/cyclonedx-api-2.0.schema.json` + all 34 files in `schema/2.0/model/`, excluding `*-bundled*` files. All `$ref`s followed transitively; every file metaschema-validated against JSON Schema draft 2020-12.

---

## 1. Broken / fragile references

All **intra-`schema/2.0` refs resolve correctly** (file + `#/$defs/...` pointer) — 0 hard-broken refs among the 36 files. However, five refs escape the versioned tree via `../../` and two of them have a real identifier mismatch:

| # | File / JSON pointer | Ref | Issue | Suggested fix |
|---|---|---|---|---|
| 1.1 | `model/cyclonedx-cryptography-2.0.schema.json` `#/$defs/cryptoProperties/properties/algorithmProperties/properties/algorithmFamily` | `../../cryptography-defs.schema.json#/definitions/algorithmFamiliesEnum` | Target file exists and pointer exists, **but** the ref resolves (per RFC 3986 against the base `$id`) to `https://cyclonedx.org/schema/cryptography-defs.schema.json`, while that file declares `"$id": "http://cyclonedx.org/schema/cryptography-defs.schema.json"` (http vs https). `$id`-based resolvers/registries will fail to match. Also a draft-07 file (`definitions`, not `$defs`) referenced from a 2020-12 schema, and an unversioned dependency outside `schema/2.0/`. | Update the external file's `$id` to `https://` (or ref the `http://` URI explicitly); long-term, snapshot a 2020-12 copy under `schema/2.0/`. |
| 1.2 | same file, `.../properties/ellipticCurve` | `../../cryptography-defs.schema.json#/definitions/ellipticCurvesEnum` | Same as 1.1. | Same as 1.1. |
| 1.3 | `model/cyclonedx-license-2.0.schema.json` `#/$defs/license/properties/id` | `../../spdx.schema.json` | Target exists, but resolves to `https://cyclonedx.org/schema/spdx.schema.json` while the file declares `"$id": "http://cyclonedx.org/schema/spdx.schema.json"` — same http/https mismatch; draft-07 target. | Same as 1.1. |
| 1.4 | `model/cyclonedx-behavior-2.0.schema.json` `#/$defs/behaviorInstance/properties/behavior` and `#/$defs/behaviorNode/properties/behavior` | `../../behavior-taxonomy.schema.json` | Resolves cleanly (`$id` is `https://…`, draft 2020-12) ✅ — but it is an **unversioned** dependency outside `schema/2.0/`; future edits to the shared file silently change 2.0 semantics. | Consider versioning it under `schema/2.0/` or pinning. |

## 2. `$id` consistency — ✅ (within 2.0)

Every one of the 36 files has `$id` exactly matching `https://cyclonedx.org/schema/2.0/<path>` and its actual path/filename, and `$schema: https://json-schema.org/draft/2020-12/schema`. The only `$id` problems are the external http/https mismatches in 1.1–1.3.

## 3. Orphaned files / definitions

- **Files:** none orphaned — all 34 model files are transitively reachable from the root schema. ✅
- **`cyclonedx-api-2.0.schema.json` is an empty stub**: `{"type": "object", "$defs": {}}` — no properties, no refs. The README describes it as "the normative API-focused schema [that] reuses CycloneDX models", which it currently does not.
  *Fix:* implement it or mark it as a placeholder in the README.
- **Orphaned `$defs` (referenced from nowhere, including their own file):**

| Location | Issue | Suggested fix |
|---|---|---|
| `model/cyclonedx-common-2.0.schema.json#/$defs/organizationalEntityOrContact` | Literally `{"$comment": "TODO"}` — an unfinished stub. | Implement or delete. |
| `model/cyclonedx-common-2.0.schema.json#/$defs/timestamp` | Unused, while **48 places** across the corpus inline `"format": "date-time"` **without a `pattern`**. In 2020-12, `format` is annotation-only by default, so those 48 timestamps are effectively unvalidated, and the one def that does validate (format + pattern) is dead. | Ref `common#/$defs/timestamp` everywhere (or delete it and accept format-only). |
| `model/cyclonedx-common-2.0.schema.json#/$defs/base64` | Unused; `common#/$defs/attachment` declares `encoding: enum ["base64"]` but never validates `content` against this pattern. | Wire it into `attachment` (e.g. via if/then on `encoding`) or delete. |
| `model/cyclonedx-common-2.0.schema.json#/$defs/baseObject` | The "mixin base object for all CycloneDX entities" (composes `extensibleProperties`) is **dead code** — nothing refs it. Consequence: only the document **root** composes `extensibleProperties`; every other object is closed with `additionalProperties: false`, so the `ext:<namespace>:<name>` extension mechanism is unusable anywhere except the BOM top level. | Either have entity defs compose `baseObject`/`extensibleProperties` + `unevaluatedProperties: false`, or delete `baseObject` and document that extensions are root-only. |

## 4. Circular references & duplicate definitions

- **File-level cycle:** `cyclonedx-component-2.0.schema.json` ⇄ `cyclonedx-ai-modelcard-2.0.schema.json` (component → `modelCard`, `graphicsCollection`; ai-modelcard → `componentData`). Fully resolvable for validators — **not problematic**, but worth knowing for code generators. Self-recursive defs (nested `components`, `dependency` graph, etc.) are standard and fine.
- **Duplicate `$defs` names across files (14):** `dependency`, `flow`, `hashAlgorithm`, `identifier` (×3), `metadata`, `postalAddress`, `rating`, `requirement`, `requirements`, `risk`, `signatures`, `step`, `timestamp`, `trigger`.
  - **True concept duplicate:** `postalAddress` in `common` vs `party` — the `party` version is a superset (adds `isoCode`, `coordinates`). Two competing address models. *Fix:* keep one (party's) and ref it; delete/alias the other.
  - **Intentional but collision-prone:** `hashAlgorithm` and `timestamp` in `cyclonedx-jss_X590_2023_10` deliberately differ from `common` (IANA names / X.590 clause 6.1 millisecond rule). Fine semantically, but same-named defs across files will clash in naive bundlers/codegen. *Fix:* prefix jss defs (e.g. `jssTimestamp`) or namespace during bundling.
  - The rest (`dependency` in dependency vs requirement, `flow`, `metadata`, `identifier`, `rating`, `risk`, `step`, `trigger`, …) are genuinely different domain concepts sharing generic names — acceptable, same bundling caveat.

## 5. Draft 2020-12 correctness

- **Metaschema validation: 0 violations in all 36 files.** ✅
- **Regex patterns: all `pattern`/`patternProperties` compile.** ✅
- **Root composition (`unevaluatedProperties: false` + `allOf: [$ref → common#/$defs/extensibleProperties]`): works as intended.** In 2020-12, `unevaluatedProperties` sees annotations from `patternProperties` inside schemas reached through `allOf`/`$ref`, so `ext:…` properties are correctly allowed at the root while unknown properties are rejected. ✅
- **`additionalProperties: false` + `allOf` combinations** (e.g. `component#/$defs/component`, `behavior#/$defs/behaviorGraph|behaviorNode`, `vulnerability#/$defs/presenceEvidence`, `threat` categories): verified that **no** `allOf` branch introduces property names absent from the parent's `properties` — all branches are `not`/`if-then` constraints on declared properties, so nothing is falsely rejected. ✅ However, this closes those objects to `ext:` extensions (see §3 `baseObject`) and mixes two closure idioms (`additionalProperties: false` vs the root's `unevaluatedProperties: false`). *Suggested fix:* standardise on one idiom per the intended extensibility model.
- **`$schema` declarations:** present and correct in every file. ✅ No legacy keywords (`definitions`, boolean `exclusiveMinimum`, `additionalItems`) anywhere. ✅ (`cyclonedx-definition-2.0.schema.json#/$defs/definitions` is a def *named* "definitions", not the keyword — legal, mildly confusing.)
- **`specVersion` is unconstrained:** `cyclonedx-2.0.schema.json#/properties/specVersion` is required but has only `examples: ["2.0"]` — any string validates. 1.x schemas pinned it with `enum`. *Fix:* add `"const": "2.0"` (or an enum/pattern).
- **Single-value enums (prefer `const`):** `cyclonedx-2.0.schema.json#/properties/specFormat` (`["CycloneDX"]`), `ai-modelcard#/$defs/energyMeasure/properties/unit` (`["kWh"]`), `ai-modelcard#/$defs/co2Measure/properties/unit` (`["tCO2eq"]`), `common#/$defs/attachment/properties/encoding` (`["base64"]`), `evidence#/$defs/assessmentMethod/allOf/1/.../not` (`["filename"]`). Cosmetic.
- **`jss…#/$defs/signatureObject` sets `unevaluatedProperties: true`:** redundant (open is the default) but intentional and documented via `$comment` citing X.590 clause 6.2. OK as-is.
- **Non-standard `meta:enum` keyword:** used consistently and every `meta:enum` key set exactly matches its sibling `enum` values ✅ (established CycloneDX 1.6/1.7 convention).

## 6. Naming / structural inconsistencies

| Location | Issue | Suggested fix |
|---|---|---|
| `model/cyclonedx-jss_X590_2023_10-2.0.schema.json` | **Not a typo:** ITU‑T **X.590** (2023‑10) *is* the JSON Signature Scheme (JSS) — distinct from X.509; `common#/$defs/signatures` even links the ITU X.590 publication. Only the *style* is inconsistent: underscores + embedded date vs the hyphenated, undated names of every other file. | Optionally rename to e.g. `cyclonedx-jss-x590-2.0.schema.json` for consistency. |
| jss `signatureObject` properties | `snake_case` (`hash_algorithm`, `public_key`, `cert_url`, `valid_from`, …) vs camelCase everywhere else — intentional (X.590 wire format), but worth a `$comment` if not already obvious. | Document; keep as-is. |
| Root container shape inconsistency | `components`, `dependencies`, `compositions`, `vulnerabilities`, `controls`, `annotations`, `formulation`, `blueprints`, `citations`, `perspectives` are **arrays**, but `threats`, `risks`, `declarations`, `definitions`, `profiles` are **objects** wrapping same-named arrays (`bom.threats.threats[]`, `bom.risks.risks[]`). | Accept (matches 1.6 `declarations` precedent) or flatten for uniformity; at minimum document the convention. |
| 26 of 307 `$defs` lack `title` | e.g. `common#/$defs/refType`, `refLinkType`, `component#/$defs/version`, `versionRange`, `componentData`, `composition#/$defs/aggregateType`, `formulation#/$defs/step`, `command`, … | Add titles for doc-generation consistency. |
| `schema/2.0/README.md` | Table references `cyclonedx-api-2.0-bundled.schema.json`, which **does not exist**. | Create it or remove the row. |
| `schema/2.0/model/README.md` | Badly stale: lists **4 nonexistent files** (`cyclonedx-ai-model-parameters`, `cyclonedx-licensing`, `cyclonedx-patent-assertion`, `cyclonedx-patent-family`) and **omits ~16 real files** (behavior, blueprint, business-objective, certification, citation, control, data, evidence, jss, party, perspective, physical, profile, requirement, risk, threat, usecase, weakness). | Regenerate the model index. |
| `component#/$defs/component/properties/mime-type` | Kebab-case like legacy `bom-ref`; everything else camelCase. | Accept as 1.x-compat or rename (`mimeType`) since 2.0 is a breaking release. |

## 7. Explicitly checked and found fine

- ✅ All 36 files parse as valid JSON and validate against the draft 2020-12 metaschema (0 errors).
- ✅ All intra-`schema/2.0` `$ref`s (file + `#/$defs/...` pointer) resolve; all pointers into external files (`#/definitions/...`) also exist.
- ✅ `$id` ↔ path/filename convention correct in all 36 files; `$schema` present everywhere.
- ✅ No orphaned model *files* — every model file is reachable from the root schema.
- ✅ Root `unevaluatedProperties: false` + `allOf → extensibleProperties` composition behaves correctly under 2020-12 annotation rules.
- ✅ No `additionalProperties: false` object is broken by its `allOf` branches (no branch declares undeclared property names).
- ✅ All regexes compile; `serialNumber` urn:uuid pattern is correct; `patternProperties` `^ext:...$` anchored correctly.
- ✅ No `required` entries missing from `properties` in closed objects; no duplicate `required` entries; no unknown `format` values; no `$ref` siblings that add conflicting constraints; `meta:enum` fully consistent with `enum` everywhere; `$defs` names uniformly lowerCamelCase.
- ✅ `jss_X590` filename is *not* an X.509 typo — it correctly references ITU-T X.590 (JSS), edition 2023-10.

### Priority fixes (summary)

1. Fix `http` vs `https` `$id` mismatch in `schema/cryptography-defs.schema.json` and `schema/spdx.schema.json` (breaks `$id`-based ref resolution from 2.0).
2. Resolve the dead extensibility design: orphaned `baseObject` + closed objects means `ext:` properties only work at the BOM root.
3. Implement or clearly stub `cyclonedx-api-2.0.schema.json`; add the missing `cyclonedx-api-2.0-bundled.schema.json` or fix the README.
4. Finish or remove `common#/$defs/organizationalEntityOrContact` (`TODO` stub); wire up or remove orphaned `timestamp`/`base64` defs (and add `pattern` to the 48 unvalidated `date-time` fields if strict validation is intended).
5. Constrain `specVersion` (e.g. `const: "2.0"`).
6. Deduplicate `postalAddress` (common vs party); regenerate `model/README.md`.

----


actions taken:
1. will not do. the schema has an existing and working and in-use ID
2. might consider - not in the scope of this PR
3. out of scope for this PR
4. out of scope of this PR
5. not planned. it is a free string that needs to stay like this for compatibility reasons with future/previous releases.
6. - `postalAddress`?? need to look into this - porobably out of the scope of this PR
   - out odscope - see https://github.com/CycloneDX/specification/issues/1035

taleodor:
2 issues flagged by AI (please check if this is valid and relevant, happy to open separate issues if needed):

1. The party model's contact email is now a rich array ("email": [{ "address": "..." }] - this PR migrates fixtures to that), while organizationalContact used by metadata.authors keeps the plain "email": "..." string.
2. SBOM metadata still uses manufacturer and author[] fields, shouldn't those be migrated to the party model?

**Edit**: Moved to #1037 following comments below.

jkowalleck:
> 2 issues flagged by AI (please check if this is valid and relevant, happy to open separate issues if needed):
> 
> 1. The party model's contact email is now a rich array ("email": [{ "address": "..." }] - this PR migrates fixtures to that), while organizationalContact used by metadata.authors keeps the plain "email": "..." string.
> 2. SBOM metadata still uses manufacturer and author[] fields, shouldn't those be migrated to the party model?

this PR is for fixes, not for changes.
Please open dedicated tickets to track these findings.

---

continues in https://github.com/CycloneDX/specification/issues/1037

jkowalleck:
re: https://github.com/CycloneDX/specification/pull/1001#issuecomment-5372315361

@taleodor Please continue reviewing this, your effort is very much appreciated

PS: continued in https://github.com/CycloneDX/specification/issues/1037

tools/src/test/resources/2.0/invalid-component-choice-missing-name-2.0.json
jkowalleck: this was intended, agreed by @stevespringett

PR #1010, opened 2026-08-07
CBOM: add protocol registry schema

bhess:
## Summary

This adds a protocol registry to `cryptography-defs`, alongside the existing algorithm families and elliptic curves. Protocols are defined per family and version, each with its algorithm composition, so tools can answer questions like "can this protocol version negotiate a post-quantum key exchange?" without every CBOM re-describing the protocol.

The TLS 1.3 entry is illustrative for now; the list will be populated further (TLS 1.2, SSH, IKEv2, ...).

## What's added

A new top-level `protocols` array. Each protocol version carries a `composition`: a list of functional slots that all apply together.

- `role`: what the slot does (`key-exchange`, `signature`, ...); open string with examples, like `variant.primitive`.
- `selection`: how many algorithms of the set are used at runtime: `all-of`, `any-of`, or `one-of`.
- `selectedBy`: how that choice is made: `negotiation`, `configuration`, `server-selected`, ...
- `algorithmSet`: algorithm names matching the variant patterns of the algorithm registry, or named bundles `{name, algorithms[]}` for cipher suites and hybrid key exchange groups.

Example, TLS 1.3 key exchange:

```json
{
  "role": "key-exchange",
  "selection": "one-of",
  "selectedBy": "negotiation",
  "algorithmSet": [
    "x25519",
    { "name": "X25519MLKEM768", "algorithms": ["x25519", "ML-KEM-768"] }
  ]
}
```

## Backwards compatibility

No breaking changes, `protocols` is optional. `protocolFamiliesEnum` uses the same tokens as `protocolProperties.type`, so a BOM references a registry entry via `protocolProperties.type` plus `version`.. no new field in the core schema needed.


schema/cryptography-defs.schema.json
```diff
@@ -229,6 +229,76 @@
...
+            "type": "array",
+            "title": "Versions",
+            "description": "List of versions of the protocol family, each with its algorithm composition.",
+            "items": {
+              "type": "object",
+              "title": "Protocol Version",
+              "description": "Defines a specific protocol version and its algorithm composition.",
+              "additionalProperties": false,
+              "properties": {
+                "version": {
```
Mehrn0ush, line 271: How should tools join this to a BOM? Is the lookup an exact match on protocolProperties.type + protocolProperties.version → family + this version string (e.g. "1.3" only)? Worth documenting, since there’s no schema $ref or uniqueness constraint on the pair.

schema/cryptography-defs.schema.json
```diff
@@ -587,6 +657,152 @@
...
+          "description": "The set of algorithms that can fill this slot."
+        }
+      },
+      "required": [
+        "role",
+        "selection",
+        "algorithmSet"
+      ]
+    },
+    "algorithmSet": {
```
Mehrn0ush, line 787: Pattern matching isn’t enforced by JSON Schema here — only plain strings. right? For PQ-readiness queries, should there be a short resolution rule (or CI check) so bad names don’t silently fail lookup?

schema/cryptography-defs.schema.json
```diff
@@ -587,6 +657,152 @@
...
+            "unknown"
+          ]
+        },
+        "algorithmSet": {
+          "$ref": "#/definitions/algorithmSet",
+          "title": "Algorithm Set",
+          "description": "The set of algorithms that can fill this slot."
+        }
+      },
+      "required": [
```
Mehrn0ush, line 781: selectedBy is in the PR description and all TLS examples, but it’s not required here. Intentional, or should it be required when selection is one-of / any-of?

Mehrn0ush (commented):
Hi, 
The composition model and TLS 1.3 example make sense to me. I left a few inline questions on join semantics, pattern resolution, and selectedBy; all minor while this is draft. 
One more thing I didn’t inline is:  the registry looks like “what a protocol version can negotiate,” while existing protocolProperties.cipherSuites / IKEv2 fields describe this instance. A short note on that layering (and how registry names relate to IANA-style identifiers in CBOMs) would make the split clearer for implementers.

schema/2.0/model/cyclonedx-cryptography-2.0.schema.json
jkowalleck, line 1: could you add new valid/invalid test cases, if needed?

schema/cryptography-defs.schema.json
bhess: Yes, that's the idea: protocolProperties.type == family and protocolProperties.version == version. This is now documented on both version descriptions, together with the rule that family/version pairs shall be unique.

schema/cryptography-defs.schema.json
bhess: If you refer to the algorithm names in the set: right, JSON Schema only sees plain strings and can't enforce the pattern match. The rule is now stated in the descriptions ("a name shall match at least one variant pattern; a name matching no pattern does not resolve").

schema/cryptography-defs.schema.json
bhess: The mechanism might be unknown at the registry level, so omission means "unstated"; the field has meanwhile been refined into selectedBy (mechanism only, including a new protocol-fixed for selections pinned by the specification itself) plus an open selectingParty, so every positive claim has an explicit value.

schema/2.0/model/cyclonedx-cryptography-2.0.schema.json
bhess: Added.. one simple positive and a negative test case.

Mehrn0ush:
One small thing I'd still like to add is a note saying that the registry is describing what a protocol version can support/negotiate and that the protocolProperties.cipherSuites / IKEv2 fields are describing this BOM instance.

It might also be worth saying something about how bundle names here are intended to map to IANA-style identifiers used in CBOMs.

Other than that, I think the join docs and the selectedBy/selectingParty split look good to me.

bhess:
> One small thing I'd still like to add is a note saying that the registry is describing what a protocol version can support/negotiate and that the protocolProperties.cipherSuites / IKEv2 fields are describing this BOM instance.
> 
> It might also be worth saying something about how bundle names here are intended to map to IANA-style identifiers used in CBOMs.
> 
> Other than that, I think the join docs and the selectedBy/selectingParty split look good to me.

Thank you, and good points. I've added a commit clarifying some of the descriptions.

tools/src/test/resources/2.0/valid-cryptography-protocol-type-2.0.json
```diff
@@ -0,0 +1,21 @@
+{
+  "$schema": "https://cyclonedx.org/schema/2.0/cyclonedx-2.0.schema.json",
+  "specFormat": "CycloneDX",
+  "specVersion": "2.0",
+  "version": 1,
+  "serialNumber": "urn:uuid:3e671687-395b-41f5-a30f-a58921a69b7a",
+  "components": [
```
bhess, line 7: wrap components in a definitions object

tools/src/test/resources/2.0/invalid-cryptography-protocol-type-2.0.json
```diff
@@ -0,0 +1,21 @@
+{
+  "$schema": "https://cyclonedx.org/schema/2.0/cyclonedx-2.0.schema.json",
+  "specFormat": "CycloneDX",
+  "specVersion": "2.0",
+  "version": 1,
+  "serialNumber": "urn:uuid:3e671687-395b-41f5-a30f-a58921a69b7b",
+  "components": [
```
bhess, line 7: wrap components in a definitions object

schema/2.0/model/cyclonedx-cryptography-2.0.schema.json
```diff
@@ -754,46 +754,14 @@
           "additionalProperties": false,
           "properties": {
             "type": {
-              "type": "string",
+              "$ref": "../../cryptography-defs.schema.json#/definitions/protocolFamiliesEnum",
               "title": "Type",
```
bhess, line 758: Rename: "Protocol type"

jkowalleck:
just merged in latest base branch(`2.0-dev`), which adds some QA tools for the non-core schemas, like `cryptography-defs`.


this PR introduces some new issues.
see https://github.com/CycloneDX/specification/actions/runs/34492663259/job/102923108162?pr=1010
and the issues/findings: https://github.com/CycloneDX/specification/actions/runs/34492663259/artifacts/10158402600

bhess:
> just merged in latest base branch(`2.0-dev`), which adds some QA tools for the non-core schemas, like `cryptography-defs`.
> 
> this PR introduces some new issues. see https://github.com/CycloneDX/specification/actions/runs/34492663259/job/102923108162?pr=1010 and the issues/findings: https://github.com/CycloneDX/specification/actions/runs/34492663259/artifacts/10158402600

Thanks! I'll have a look.

tools/src/test/resources/2.0/invalid-cryptography-protocol-type-2.0.json
bhess: I have checked this, but it seems that "components" are not yet implemented in the 2.0 definitions schema: https://github.com/CycloneDX/specification/blob/2.0-dev/schema/2.0/model/cyclonedx-definition-2.0.schema.json.

stevespringett:
This change does not make any normative changes to the Ecma standard. The implementation has been changed to externalize protocols into the cryptography-def subschema. Because this change is only an editorial update to the standard, I am merging it without the RFC period for TC54.

PR #1017, opened 2026-08-10
2.0 dev split docsgen

jkowalleck:
move those docgen for schema-v1 to own dir, tp path the way for new docs tools


jkowalleck:
example run https://github.com/CycloneDX/specification/actions/runs/31417245320 
res
- https://github.com/CycloneDX/specification/actions/runs/31417245320/artifacts/9073895045
- https://github.com/CycloneDX/specification/actions/runs/31417245320/artifacts/9073886842
- https://github.com/CycloneDX/specification/actions/runs/31417245320/artifacts/9073899087

Issue #954, opened 2026-06-19
Proposal: Safety Relevance Metadata for CycloneDX


devashridatta-dotcom:
## Summary

This proposal introduces an optional **Safety Relevance** metadata property for CycloneDX components.

The goal is to capture whether a component participates in safety-critical or mission-critical functions where compromise, malfunction, or misuse could contribute to:

* Physical harm
* Human safety risks
* Environmental damage
* Critical infrastructure disruption
* Societal impact

The proposal complements existing CycloneDX capabilities such as Vulnerability Disclosure, VEX, AI BOM, and dependency analysis by adding contextual information about the operational consequences of component failure.

## Motivation

CycloneDX provides rich metadata describing:

* Component identity
* Dependencies
* Vulnerabilities
* VEX assertions
* Services
* AI and machine learning artifacts

However, it does not currently express whether a component is part of a safety-critical system.

Organizations in automotive, healthcare, robotics, industrial control systems, energy, and AI increasingly require this context to:

* Prioritize remediation
* Interpret VEX results
* Build safety cases
* Meet regulatory obligations
* Improve risk governance

## Initial Proposal

Introduce optional metadata properties:

* cyclonedx:safety-relevance
* cyclonedx:safety-domain
* cyclonedx:safety-impact

Example:

safety-relevance = Critical

safety-domain = Medical

safety-impact = HumanSafety

## Example Classification

* None
* Indirect
* Critical
* Unknown

## Questions for the Community

1. Should Safety Relevance begin as custom properties or as a first-class schema object?
2. Are there existing use cases in medical, automotive, AI, or industrial systems?
3. How should Safety Relevance interact with VEX and vulnerability analysis?
4. Would a common taxonomy benefit downstream tooling and SBOM consumers?

## Next Steps

If there is interest, I can:

* Draft a formal schema proposal
* Provide example BOMs
* Develop mappings to VEX and AI BOM
* Collaborate with the community on taxonomy and governance

stevespringett:
Thanks for this proposal, @devashridatta-dotcom. Safety relevance is squarely in scope for where CycloneDX is heading, and a good deal of what you describe is already taking shape in the 2.0 line (in development), so this is well-timed. Here is where each part maps today and where the gap is.

### What already exists in 2.0

[...]

PR #1041, opened 2026-08-24
Add safety integrity level schema support

devashridatta-dotcom:
## Summary

Adds the small first-class safety-integrity classification discussed in #954 for the CycloneDX 2.0 schema line.

This PR is intentionally narrow. It does not try to move the full SRAP vocabulary into CycloneDX core. Instead, it implements the piece that appears to be missing from the existing 2.0 model: a standards-aligned, machine-readable way to express safety integrity classification in context.

## What changes

- Adds a shared `safetyIntegrityLevel` schema definition with scheme-gated levels for `ISO-26262`, `IEC-61508`, `DO-178C`, `IEC-62304`, and `EN-50128`, plus a custom scheme option.
- Allows `safetyIntegrityLevels` on blueprint asset classifications and requirements, keeping the assertion scoped to a system/context carrier rather than treating it as an intrinsic component property.
- Adds `hazard-analysis` and `safety-case` external reference types for supporting safety evidence.
- Adds `EPSS` as a vulnerability score method.
- Adds `safety` as a requirement type so safety requirements can be identified without relying only on free text.

## Why this is the narrow CycloneDX-native path

The discussion in #954 clarified that CycloneDX 2.0 already has first-class structures for most SRAP needs:

- `risks` for safety impact and risk linkage.
- `perspectives` and blueprints for system/context views.
- `criticality` for contextual asset importance.
- `vulnerabilities` and VEX analysis for vulnerability impact.
- `declarations`, `claims`, `attestations`, and `evidence` for assertion and evidence chains.
- `definitions.requirements` for safety requirements.

The remaining core gap is safety integrity classification: ASIL, SIL, DAL, IEC 62304 software safety class, and similar scheme-specific levels. This PR focuses on that gap.

## Relationship to SRAP

SRAP-style concepts such as `SafetyContextRelationship`, `ChangeTrigger`, `ChangeImpactAnalysis`, `RequirementVerification`, and `EvidenceRelationship` can be represented today using existing CycloneDX structures and, where needed, profile-owned terms such as `srap:*` properties or structured `ext:*` extensions.

Those broader SRAP terms should remain in the profile layer unless TC54 decides that any of them belong in CycloneDX core. This PR keeps the CycloneDX schema change small while preserving alignment with SPDX safety work.

## Automotive/Product Security and Risk context

This also aligns with the emerging Product Security and Risk Working Group direction. Automotive is a good first use case because ISO 26262 and ASIL are common safety integrity concepts, and ASRG's move toward CycloneDX for automotive risk representation makes a consistent safety integrity vocabulary useful for downstream tooling and policy engines.

## Validation

- Parsed all touched JSON schema files successfully.
- `pnpm run test:v2.0:t2-json-schema-semantic` passes.
- Targeted AJV checks for `safetyIntegrityLevel` pass:
  - `ISO-26262` + `asil-d` accepted.
  - `ISO-26262` + `sil-3` rejected.
  - `IEC-62304` + `class-c` accepted.
  - Custom scheme object accepted.

Note: `pnpm run test:v2.0:t1-json-schema-validate` and the functional test currently fail on a clean upstream `2.0-dev` checkout with the same unresolved reference:

`model/cyclonedx-common-2.0.schema.json#/$defs/extensibleProperties`

So that failure appears inherited from the current `2.0-dev` test wiring rather than introduced by this change.

## RFC question

Is this narrow `safetyIntegrityLevels[]` change the right candidate to take through the 30-day RFC period, with the broader SRAP vocabulary handled as profile guidance or a companion schema?


schema/2.0/model/cyclonedx-common-2.0.schema.json
```diff
@@ -950,6 +954,225 @@
       "$ref": "cyclonedx-jss_X590_2023_10-2.0.schema.json#/$defs/signatures",
       "title": "Signatures",
       "description": "Enveloped signatures in [JSON Signature Scheme (JSS/ITU-T X.590)](https://www.itu.int/epublications/publication/itu-t-x-590-2023-10-json-signature-scheme-jss)."
+    },
+    "safetyIntegrityLevel": {
```
jkowalleck: could this be moved into `cyclonedx-behavior-2.0.schema.json`

maybe it is abetter fit.

devashridatta-dotcom: Moved safetyIntegrityLevel into cyclonedx-behavior-2.0.schema.json and updated the blueprint/requirement references to point there. Thanks for the direction.

devashridatta-dotcom:
Thank you for syncing this branch with the latest `2.0-dev`.

Could maintainers advise on the next expected step for PR #1041?

Specifically, should this narrow `safetyIntegrityLevels[]` proposal proceed toward the 30-day RFC process for CycloneDX 2.0, or would you prefer changes to the schema shape, naming, scope, or placement before RFC consideration?

I am happy to revise quickly based on maintainer guidance.

Issue #1066, opened 2026-08-29
[2.0]: Proposal - Pre-defined perspectives: stable spec-provided identity with incorporation by reference


taleodor:
## Describe the feature

Perspectives (#736, #745) are fully author-defined: identity is carried only by the free-text name. This leaves two gaps:

1. No stable identity. Tooling cannot reliably recognize that a document carries, say, a model card view, since matching on free-text names is fragile (case, wording, localization). This matters more now that #990 points model cards at perspectives: "Model Card" is an industry-adopted term, and consumers need a stable way to find that view.
2. No reference form. A well-known perspective must be restated in full in every document that wants it. 

## Possible solutions

Add an optional predefined field to perspective:

```
{
  "perspectives": [
    { "predefined": "model-card" }
  ]
}
```

## Alternatives

Considered:
- Reserved name values (registry-governed, like the property taxonomy): no schema change and registry-paced growth, but the schema cannot express "mappings optional iff the name is reserved" without embedding the reserved list, so the reference form is impossible — and it overloads display text as machine identity.
- Identity field on inline perspectives (tagging inline mappings with a well-known identity): rejected — nothing could validate that the inline content faithfully represents the claimed identity, inviting stripped-down or contradictory "model cards."

## Additional context
References:
#736 / #745 — perspectives concept and model
#990 — model cards expressed via perspectives (this proposal closes the loop that PR opens)
#960 — perspectives/ catalog precedent (PQC readiness)

PR #1067, opened 2026-08-29
Introduce pre-defined perspectives

taleodor:
- Implements #1066.

**What this adds**

**1. Schema: three forms of a perspective, and `scopes` (cyclonedx-perspective-2.0.schema.json)**

A perspective is one of three mutually exclusive forms (oneOf, enforced in all directions):

```json
{ "bom-ref": "card-vision", "predefined": "cdx:perspectives:model-card", "predefinedVersion": 1, "scopes": [ … ] }
{ "bom-ref": "crypto-per-model", "ref": "def-crypto-in-use", "scopes": [ … ] }
{ "bom-ref": "custom", "name": "…", "mappings": [ … ], "scopes": [ … ] }
```

- **Pre-defined form** (`predefined` + `predefinedVersion`): incorporates a published perspective from the CycloneDX perspectives catalog by identity, giving tooling stable spec-backed identity instead of free-text name matching. Identities live under the reserved `cdx:` namespace with a `cdx:perspectives:` path (per @mrutkows), so they cannot collide with author-chosen names and can be declared in the reserved cyclonedx-property-taxonomy. `predefinedVersion` pins the published revision: it is the `version` of the catalog document defining the perspective (integer, incremented by 1 per revision, like other CycloneDX version fields). Initial identities: `cdx:perspectives:model-card` and `cdx:perspectives:pqc-readiness` (the latter reserved for the entry proposed in #960).
- **Definition reference form** (`ref`): applies a reusable inline perspective declared under `definitions.perspectives` (see 1b), by bom-ref or BOM-Link element.
- **Inline form** (`name` + `mappings`): fully defined in the document.

In the two reference forms only `bom-ref` and `scopes` may accompany the reference, so the referenced definition stays the single source of truth while the document decides where it applies.

**Why a field rather than the choice pattern.** A field on the perspective object keeps a pre-defined perspective addressable. A bare enum string cannot carry a bom-ref, so it could never be the target of a BOM-Link or any intra-document reference. With the field form, `{ "bom-ref": "…", "predefined": "cdx:perspectives:model-card", "predefinedVersion": 1 }` remains a first-class, referenceable node, and the oneOf enforces the same exclusivity the choice pattern would have.

**`scopes`: applying a perspective to parts of the document.** Without scopes a perspective applies to the whole document. With scopes, each entry is an independent application: one view, one completeness evaluation.

```json
"scopes": [
  { "bom-ref": "scope-vision", "name": "Vision model", "targets": ["model-vision"] },
  { "bom-ref": "scope-text",   "name": "Text models",
    "expressions": ["$.definitions.components[?(@.type=='machine-learning-model' && @.group=='text')]"] }
]
```

- `expressions[]`: JSONPath (RFC 9535) from the document root; `targets[]`: bom-ref or BOM-Link (`elementLink`, the type compositions use). At least one of the two; within a scope they union.
- Resolution (normative text on `scopes`, written as a numbered algorithm so implementations converge): (1) evaluate expressions from the root; (2) resolve targets by bom-ref; (3) one-hop closure: every same-document reference entry (an object with a `ref` property, as the inventory refactor introduces) inside the selected subtrees also selects the referenced node; (4) the scope is the union of the selected subtrees. Containment is by JSONPath Normalized Path prefix (RFC 9535 §2.7). BOM-Links to other documents are not followed.
- Mapping expressions stay root-absolute: a mapping selects the intersection of its expression with the scope. Nothing is relative, so catalog entries stay generic and the document decides where they apply. Published pre-defined perspectives and definitions declare no scopes of their own.
- Closure is tooling behaviour; the schema validates structure only.

**`via` on a mapping: following references.** JSONPath cannot join, so data *related* to a scoped subject (the datasets a model trained on, the risks that affect it, the algorithms a library provides) needed a way to be reached. `via` is an ordered list of traversal steps applied before the expression; each step is exactly one of:

- `{ "refs": "<root-absolute JSONPath>" }`: restricted to the current node set, collect reference values and resolve them to objects in this document. Forward references, e.g. a model's `training.datasets`.
- `{ "referrers": "<root-absolute JSONPath>" }`: the inverse. The path selects reference values anywhere in the document; the step produces the objects that hold a value identifying a node in the current set (the object owning the property, or, when the value sits in an array, the object holding the array). For relationships that point *towards* the scoped nodes, e.g. a risk's `affects`. The step replaces the current set with the matching referring objects; when the current set is the whole document it produces the whole document, so an unscoped perspective is not narrowed.
- `{ "dependencies": "depends-on" | "provides" | "dependents" | "provided-by", "transitive": bool }`: walk the dependency graph from the referenceable nodes within the current set.

The first step starts from the scope (or the whole document when unscoped); the expression is then intersected with the subtrees of what the last step produced. Steps chain:

```json
{ "expression": "$.definitions.components[?(@.type=='cryptographic-asset')]",
  "via": [ { "dependencies": "depends-on", "transitive": true }, { "dependencies": "provides" } ] }
```

Worked example for `referrers` (from @JunHwan-Kwon's review): two models, `risk-a` with `"affects": ["model-a"]`, `risk-b` with `"affects": ["model-b"]`. A model card scoped to `model-a` evaluates a risk mapping as: current set `{model-a}` → `referrers: $.risks.risks[*].affects[*]` yields `{risk-a}` → expression intersected with that subtree → `risk-a`. Unscoped, both risks are selected. Selection is automatic; authors do not add risks to the scope.

**1a. Registry: `schema/perspectives-defs.schema.json` + `schema/perspectives-defs.json`**

The identity enum does not live in the versioned perspective schema. It lives in a standalone registry at the schema root, which the perspective schema `$ref`s (`../../perspectives-defs.schema.json#/definitions/preDefinedPerspectivesEnum`), following the `cryptography-defs` pattern, so perspectives can be added or revised without a specification release.

- `perspectives-defs.schema.json` is hand-maintained: the identity enum (with `meta:enum` descriptions) plus the governing schema for the registry data. No mapping is declared; by naming convention the catalog document defining `cdx:perspectives:<name>` is `perspectives/<name>-perspective.json`.
- `perspectives-defs.json` is generated. Per identity it records every published version with the sha256 of the catalog document's canonical JSON and the commit that introduced it. Once registered, a version's content is immutable; changing a catalog document requires bumping its `version` by 1. Identities without a catalog document yet (`pqc-readiness`) are reserved and absent until their document lands.
- CI: the schema-v2 suite checks every identity on each PR (reserved, new, unchanged, or bumped by exactly one pass; a content change without a bump, a regressed or skipped version, or a removed registered document fail). A workflow runs the generator after merge and opens a PR registering any new version, mirroring `generate_algorithm_families.yml`.

**1b. Reusable definitions: `definitions.perspectives` (cyclonedx-definition-2.0.schema.json)**

Reusable inline perspectives, each with a required bom-ref, typed as `perspectiveDefinition`, a mixin over the perspective schema that forbids `predefined`, `predefinedVersion`, `ref` and `scopes`. The same definition can be applied at several places with different scopes without repeating its mappings. Follows the definitions-registry direction of #1075.

**2. Catalog: perspectives/model-card-perspective.json**

Second entry in the perspectives/ catalog introduced by #960, following its delivery shape (a complete, minimal, valid 2.0 document containing only the perspective). This is the definition incorporated when a document declares `"predefined": "cdx:perspectives:model-card"`. 23 mappings, eight reached through `via`:

- **Model identity**: details (name, version, description), identifiers (identity claims grouped by asserting party, e.g. a Package-URL for a model hub entry), supplier via party roles, licence.
- **Intrinsic characteristics** via `modelProperties`: tasks, learning types, architecture, inputs/outputs, size and quantization, evaluation, limitations, environmental considerations. Models are located by descendant search (`$..[?(@.type=='machine-learning-model')]`) so the entry applies whether models are declared at the document root, in inventories, or in definitions.
- **Training**: the training object; the training procedure reached through `via.refs` over `modelProperties.training.formula` into `$.formulation[*]` (workflows and tasks as the training stages and the data each consumed).
- **Datasets**, reached through `via.refs` over `modelProperties.training.datasets`, in four mappings so completeness can report which aspect is missing: identity and licensing (required), description and contents, classification and sensitive data, governance (parties accountable as owner, steward, custodian).
- **Intended use**: the model's use-case references, and the use-case definitions reached through `via.refs`; intended users via party roles.
- **Ethical and fairness considerations** as risk model entries (ethical domain; fairness/bias impact categories), reached through `via.referrers` over `$.risks.risks[*].affects[*]`: for a scoped model, exactly the risks that name it in `affects`; unscoped, every matching risk.

Not mappable today: provenance labels such as crawled or synthetic, and data subsets with rationale; the 2.0 data model has no fields for them.

**3. Tests**
- Fixtures (auto-discovered by the schema-v2 suite). Valid: the pre-defined reference form alongside a plain inline perspective; a scoped document exercising `targets`, `expressions`, a `definitions.perspectives` entry applied at two scopes by `ref`, a chained dependency traversal and a `refs` step; the two-models/two-risks document with a model card scoped to `model-a` and an inline perspective using `referrers`. Invalid: unknown identity (correct namespace, not in the enum); reference missing `predefinedVersion`; identity plus inline mappings; identity plus name and domains; scope with neither expressions nor targets; traversal step mixing `refs` and `dependencies`; traversal step mixing `referrers` and `refs`; unknown dependency direction; `ref` plus inline mappings; `ref` plus `predefined`; a definition declaring scopes; a definition without bom-ref.
- Registry test (`test:v2.0:t4-perspectives-registry`, runs in the JavaScript CI workflow on every PR): validates `perspectives-defs.json` against its governing schema, checks registered identities are in the enum with contiguous versions from 1 and conventional file paths, and assesses each identity's catalog document against the registry.
- Harness wiring: `perspectives-defs.schema.json` registered as an external schema in both JS ajv harnesses, the Java 2.x harness's classpath mapping, and the bundler's ref-exception list, mirroring `cryptography-defs.schema.json`.

**Not included**
- Bundled schemas — left to the post-merge bundle workflow, per repo convention.
- Conformance validation — the schema validates identity, exclusivity and structure; evaluating scopes, closure, traversal steps (`refs`, `referrers`, `dependencies`) and mapping expressions against a document's data remains tooling territory, as for perspectives generally. The fixtures pin the intended node selections in prose; a semantic evaluator would be a separate contribution.

**Dependencies**
- Premised on #990: the catalog entry's mappings target `modelProperties`, `training.datasets`, `training.formula` and `useCases`. The schema change and fixtures stand alone; only the catalog entry's JSONPath expressions assume #990's shape. Happy to adjust if that lands differently.
- Written with #1075 in mind: schema examples use the inventory layout, `definitions.perspectives` follows its definitions registry, and the catalog's descendant search covers both the current and the inventory layout. Fixtures keep the root `components` layout until #1075 lands.
- Complements #960: adopts its perspectives/ folder and delivery shape, and reserves `cdx:perspectives:pqc-readiness` for its entry; to be registered automatically, that document needs to live at `perspectives/pqc-readiness-perspective.json` at `version` 1.


schema/2.0/model/cyclonedx-perspective-2.0.schema.json
```diff
@@ -66,6 +94,19 @@
...
+      "type": "string",
+      "title": "Pre-Defined Perspective",
+      "description": "Identifies a well-known, pre-defined perspective published in the CycloneDX perspectives catalog, incorporating the published definition by reference and enabling tooling to recognize the perspective without matching on free-text names. A perspective declaring a pre-defined identity shall not provide any inline content other than bom-ref; a perspective defining its own content shall omit this field.",
+      "enum": [
+        "model-card",
+        "pqc-readiness"
+      ],
+      "meta:enum": {
+        "model-card": "A transparency view for machine learning models following the industry-standard model card structure, covering model details, intended use, training data and procedure, quantitative analysis, technical limitations, and ethical and environmental considerations.",
+        "pqc-readiness": "A view for assessing readiness for the post-quantum cryptography migration, covering cryptographic inventory, quantum resistance of the cryptography in use, and cryptographic agility."
```
mrutkows: Nice. Reserving "pqc-readiness", in advance, is good foresight.

perspectives/model-card-perspective.json
```diff
@@ -0,0 +1,168 @@
...
+  "$schema": "https://cyclonedx.org/schema/2.0/cyclonedx-2.0.schema.json",
+  "specFormat": "CycloneDX",
+  "specVersion": "2.0",
+  "version": 1,
+  "metadata": {
+    "timestamp": "2026-08-29T12:00:00Z"
+  },
+  "perspectives": [
+    {
+      "bom-ref": "perspective-model-card",
```
mrutkows, line 11: I would suggest we utlilize the reserved "cdx:" namespace and establish a "cdx:perspectives" namespace path (or singular "perspective"). We could build upon and declare as part of the reserved cyclonedx-property-taxonomy.

perspectives/model-card-perspective.json
```diff
@@ -0,0 +1,168 @@
...
+      "name": "Model Card",
+      "description": "Defines the data elements of a machine learning model card, following the industry-standard model card structure: model details, intended use, training data and procedure, quantitative analysis, technical limitations, and ethical, fairness, and environmental considerations. The mappings project that structure onto the CycloneDX 2.0 decomposition of the former first-class modelCard entity: intrinsic technical characteristics live in modelProperties, which may only appear on components of type machine-learning-model; training datasets are components of type data; intended use cases are use case definitions referenced from the model; and ethical and fairness considerations are entries in the document's risk model. This perspective assumes the AI/ML model properties proposed in CycloneDX/specification#990.",
```
mrutkows: Excellent. It captured all the key items from the guide.

perspectives/model-card-perspective.json
```diff
@@ -0,0 +1,168 @@
...
+      "domains": [
+        "machine-learning",
+        "artificial-intelligence",
+        "transparency",
+        "ethics"
+      ],
+      "mappings": [
+        {
+          "expression": "$.components[?(@.type=='machine-learning-model')]['name','version','description']",
+          "nativeName": "Model Details",
```
mrutkows, line 23: Not quite sure what "native" means.

perspectives/model-card-perspective.json
```diff
@@ -0,0 +1,168 @@
...
+      "description": "Defines the data elements of a machine learning model card, following the industry-standard model card structure: model details, intended use, training data and procedure, quantitative analysis, technical limitations, and ethical, fairness, and environmental considerations. The mappings project that structure onto the CycloneDX 2.0 decomposition of the former first-class modelCard entity: intrinsic technical characteristics live in modelProperties, which may only appear on components of type machine-learning-model; training datasets are components of type data; intended use cases are use case definitions referenced from the model; and ethical and fairness considerations are entries in the document's risk model. This perspective assumes the AI/ML model properties proposed in CycloneDX/specification#990.",
+      "domains": [
+        "machine-learning",
+        "artificial-intelligence",
+        "transparency",
+        "ethics"
+      ],
+      "mappings": [
+        {
+          "expression": "$.components[?(@.type=='machine-learning-model')]['name','version','description']",
```
mrutkows: Should there be a version?  Should we include any identifiers (like pURL for huggingface hub models)?

mrutkows: or should all 2.0 identifiers have a mapping? 2.0 still confuses me as we seem to have diluted identifiers we endorse, like pURL, which are lock-step with Ecma/ISO goals, but treat it the same as a made up identity (domain) as well as legacy domains we attempted to solve with pURL.

perspectives/model-card-perspective.json
```diff
@@ -0,0 +1,168 @@
...
+        {
+          "expression": "$.components[?(@.type=='machine-learning-model')].modelProperties.training",
+          "nativeName": "Training Data and Procedure",
+          "nativeDescription": "How the model was trained: the training formula and the datasets used.",
+          "relevance": "recommended",
+          "weight": 0.8,
+          "rationale": "Training data provenance is central to assessing bias, capability boundaries, and data protection obligations."
+        },
+        {
+          "expression": "$.components[?(@.type=='data')]",
```
mrutkows: Datasets are the key focus of EU AI Act and templates.  This is one of the main concerns I have with perspectives as we need to likely capture all the things we need for the top-level component (e.g., identity, licenses, version, etc.) as well as information about its data provenance (e.g., crawled, synthetic, etc.) some of which are "properties".  Plus we need to describe the "usage" of the data relative to training steps/stages (and perhaps declare data subsets and rationale).

perspectives/model-card-perspective.json
```diff
@@ -0,0 +1,168 @@
...
+        {
+          "expression": "$.components[?(@.type=='machine-learning-model')].modelProperties.limitations",
+          "nativeName": "Technical Limitations",
+          "nativeDescription": "Known technical limitations of the model, including constraints on accuracy, reasoning, scalability, and appropriate use, and relevant performance tradeoffs.",
+          "relevance": "required",
+          "weight": 0.9,
+          "rationale": "Limitations are the card's primary safeguard against use outside the model's competence."
+        },
+        {
+          "expression": "$.risks.risks[?(@.domains[?(@.type=='ethical')])]",
```
mrutkows, line 212: I get the feeling Steve may be reworking "risk" as the object includes perhaps too much... The goal is to isolate risks, harms and benefits more effectively so perhaps they can be disassociated from most of the risk object (and optionally tied/bound to risks) and include them in "considerations" (for model design).

perspectives/model-card-perspective.json
taleodor: This is part of the existing perspective model and means the domain-specific term the target audience uses for that data, so a perspective can translate CycloneDX nomenclature into a familiar vocabulary. If you want this adjusted for this case, lmk.

perspectives/model-card-perspective.json
taleodor: implemented

perspectives/model-card-perspective.json
taleodor: probably shouldn't be in scope for this PR

perspectives/model-card-perspective.json
taleodor: Added identifiers; a strict "pURL present" check is expressible as a filter on `@.scheme=='purl'` with `relevance: required` if you want to add this.

Note, that we can't rank schemes with the current approach (e.g., can't make purl preferred over others).

perspectives/model-card-perspective.json
taleodor: Added `via` traversal steps to express, so dataset mappings are reached from the scoped model through its modelProperties.training.datasets references: a card for model A only sees A's datasets.

The single mapping is also split into four so completeness can say which aspect is missing: identity and licensing (name, version, identifiers, licences; required), description and contents, classification and sensitive data, and governance (the parties accountable as owner, steward, custodian). Data usage relative to training stages is covered by a "Training Procedure" mapping that follows training.formula into the formulation, whose workflows and tasks are the stages and the data each consumed.

Two things you list are not in 2.0 data model: provenance labels such as crawled or synthetic, and data subsets with rationale. Today they could only live in `properties`. That's a gap for #990.

schema/2.0/model/cyclonedx-perspective-2.0.schema.json
```diff
@@ -78,14 +262,24 @@
         "expression": {
           "type": "string",
           "title": "Path Expression",
-          "description": "A [JSONPath](https://datatracker.ietf.org/doc/html/rfc9535) expression that identifies the types of data relevant to this perspective.",
+          "description": "A [JSONPath](https://datatracker.ietf.org/doc/html/rfc9535) expression, evaluated from the document root, that identifies the types of data relevant to this perspective. When the perspective is applied to a scope, only the identified nodes within the scope are selected; when the mapping declares traversal steps, only those within the nodes the steps produce.",
```
JunHwan-Kwon, line 265: Thanks for adding scopes and `via`. Could you clarify how the scope restriction here interacts with the document-level handling described for Ethical Considerations and Fairness Assessments?

The PR description and both catalog rationales describe document-wide risk lookup, optionally narrowed through `affects`. The schema descriptions for `mappings` and `expression` restrict selected nodes to the scope or the output of traversal steps. I could not find a corresponding document-level exception or a traversal rule for following `affects` from a scoped model back to its risks.

I checked a minimal document at `4c7c7c2` containing two models and two risks: `risk-a` affects `model-a`, and `risk-b` affects `model-b`. Both risks match the ethical and fairness mappings.

The four fixture variants pass JSON Schema validation. The harness uses `jsonpath-rfc9535` for the catalog expressions and separately applies scope containment and the alternative `affects` filter. Results, identical for both mappings:

| Scope              | Scope-restricted selection | Document-wide selection | Document-wide, then affects filter |
| ------------------ | -------------------------- | ----------------------- | ---------------------------------- |
| model-a            | none                       | risk-a, risk-b          | risk-a                             |
| model-a and risk-a | risk-a                     | risk-a, risk-b          | risk-a                             |
| no scopes          | risk-a, risk-b             | risk-a, risk-b          | risk-a, risk-b                     |

This checks node selection, not a completeness evaluator. Since Ethical Considerations is marked `required`, the distinction could affect whether a per-model view includes the relevant risk evidence.

Related to the [fixture drift I noted in #1075](https://github.com/CycloneDX/specification/pull/1075#issuecomment-5522993668), descendant search locates models across document layouts; the question here is how to select risks that reference a scoped model.

A concrete use case is a medical device BOM containing several models, where a per-model view used for regulatory documentation needs to preserve the association between each model and the risks that affect it.

For a scope targeting only `model-a`, should `risk-a` be selected automatically, or should authors explicitly include it in the scope? If automatic selection is intended, a traversal step for the inverse direction of `affects` may be a way to express it within `via`, so the schema text and the catalog rationale state the same rule.

<details>
<summary>Minimal fixture for the model-a-only case</summary>

```json
{
  "$schema": "https://cyclonedx.org/schema/2.0/cyclonedx-2.0.schema.json",
  "specFormat": "CycloneDX",
  "specVersion": "2.0",
  "version": 1,
  "components": [
    {
      "bom-ref": "model-a",
      "type": "machine-learning-model",
      "name": "Synthetic model A",
      "version": "1.0"
    },
    {
      "bom-ref": "model-b",
      "type": "machine-learning-model",
      "name": "Synthetic model B",
      "version": "1.0"
    }
  ],
  "risks": {
    "risks": [
      {
        "bom-ref": "risk-a",
        "name": "Synthetic fairness risk A",
        "statement": "Synthetic example only: model A may perform differently across groups.",
        "domains": [{"type": "ethical"}],
        "affects": ["model-a"],
        "inherentRisk": {
          "impact": {
            "level": "moderate",
            "categories": ["fairness"]
          }
        }
      },
      {
        "bom-ref": "risk-b",
        "name": "Synthetic fairness risk B",
        "statement": "Synthetic example only: model B may perform differently across groups.",
        "domains": [{"type": "ethical"}],
        "affects": ["model-b"],
        "inherentRisk": {
          "impact": {
            "level": "moderate",
            "categories": ["fairness"]
          }
        }
      }
    ]
  },
  "perspectives": [
    {
      "bom-ref": "card",
      "predefined": "cdx:perspectives:model-card",
      "predefinedVersion": 1,
      "scopes": [{"targets": ["model-a"]}]
    }
  ]
}
```

The other cases change the targets to `["model-b"]` or `["model-a", "risk-a"]`, or omit `scopes`.

</details>

Happy to contribute the fixtures and expected node selections as regression cases once the intended semantics are settled.

taleodor: Thank you @JunHwan-Kwon - looks like a real gap and I just added a possible fix for that - with `referrers` as the inverse of `refs`, please review. so the shape would be

```json
{
  "bom-ref": "risks-per-model",
  "name": "Risks per model",
  "scopes": [{ "name": "Model B", "targets": ["model-b"] }],
  "mappings": [
    {
      "expression": "$.risks.risks[*]",
      "via": [{ "referrers": "$.risks.risks[*].affects[*]" }],
      "nativeName": "Risks affecting this model",
      "relevance": "required"
    }
  ]
}
```

And selection would be automatic, authors do not need to add individual risks to scope

JunHwan-Kwon: Thank you for addressing this so quickly.

I rechecked `fdc2bd6` against the original two-model/two-risk case. With `referrers`, both model-card risk mappings now select the relevant risk for each scoped model, and both risks when unscoped.

I also ran `npm run test:v2.0` on Linux. All four groups passed, including the new `referrers` fixtures and the catalog registry check. This resolves the scoped risk-selection gap I raised.

One small documentation follow-up: updating the PR description to mention `referrers` and the automatic scoped risk selection (and the `via` count, now eight rather than six) would help keep it consistent with the schema.

schema/2.0/model/cyclonedx-perspective-2.0.schema.json
```diff
@@ -119,6 +313,90 @@
...
+          "examples": [
+            "$..[?(@.type=='machine-learning-model')].modelProperties.training.datasets[*]"
+          ]
+        },
+        "referrers": {
+          "type": "string",
+          "title": "Referrer Path",
+          "description": "A [JSONPath](https://datatracker.ietf.org/doc/html/rfc9535) expression, evaluated from the document root, whose results are reference values (bom-ref or bom-link). The step produces the objects that hold such a value identifying a node in the current set: the object owning the property, or, when the value sits in an array, the object holding the array. This is the inverse of `refs`, for relationships that point towards the scoped nodes, such as a risk's `affects`. When the current set is the whole document the step produces the whole document, so an unscoped perspective is not narrowed to referenced nodes.",
```
JunHwan-Kwon: One small wording suggestion to clarify that a scoped object is not automatically retained by the traversal:

```suggestion
          "description": "A [JSONPath](https://datatracker.ietf.org/doc/html/rfc9535) expression, evaluated from the document root, whose results are reference values (bom-ref or bom-link). The step produces the objects that hold such a value identifying a node in the current set: the object owning the property, or, when the value sits in an array, the object holding the array. This is the inverse of `refs`, for relationships that point towards the scoped nodes, such as a risk's `affects`. When the current set is the whole document the step produces the whole document, so an unscoped perspective is not narrowed to referenced nodes. Otherwise, the step replaces the current set with the matching referring objects; objects are not retained merely because they were already in the current set. For risk mappings using `affects`, scope to the affected model to select its risks.",
```

perspectives/model-card-perspective.json
```diff
@@ -0,0 +1,248 @@
...
+            }
+          ],
+          "nativeName": "Ethical Considerations",
+          "nativeDescription": "Risks in the ethical domain associated with the model, including affected parties, benefits, harms, and mitigations.",
+          "relevance": "required",
+          "weight": 0.9,
+          "rationale": "Ethical considerations are expressed as entries in the document's risk model rather than as card-local prose, gaining structured likelihood, impact, and response data. Risks reference the models they affect through `affects`, so the referrers step selects, for a scoped model, exactly the risks that name it; unscoped, every matching risk is selected."
+        },
+        {
+          "expression": "$.risks.risks[?(@.inherentRisk.impact.categories[?(@=='fairness' || @=='bias')])]",
```
JunHwan-Kwon, line 225: Small integration note for Fairness Assessments: the risk-model changes introduced in [f6d07fe](https://github.com/CycloneDX/specification/commit/f6d07fef1cf6b7342d6debf909c1b5d35977b95d) remain present at #990 head `38dfe9c`. If #990 lands with that representation, could this expression and its `nativeDescription` be aligned with it?

That change removes `bias` from the predefined impact categories while `fairness` remains. It also introduces `bias` as a weakness kind and reusable entries under `definitions.weaknesses`; risks can associate those weaknesses through `relatedWeaknesses`. This relates to the [earlier discussion about risk-model integration under Ethical Considerations](https://github.com/CycloneDX/specification/pull/1067#discussion_r3895958300).

At #1067 head `7a7d2dd`, this filter and description still treat both fairness and bias as impact categories. If #990's representation lands, the `bias` branch will no longer match a schema-valid predefined impact category, while `fairness` remains valid.

I am flagging the cross-PR alignment point only. This is not a defect in #1067 against its current schema, and I am not proposing a traversal change here.

JunHwan-Kwon:
Thanks for incorporating the wording in `7a7d2dd`.

One separate, non-blocking question about completeness tooling. My executable probe reproduces the replacement of `{model-a}` by `{risk-a}` for the existing two-model/two-risk fixture, without retaining model A in that mapping's evaluation set. This F1 node-selection trace is unchanged from `fdc2bd6`.

The PR already includes a multi-subject scope: [`valid-perspective-scopes-2.0.json`](https://github.com/CycloneDX/specification/blob/7a7d2dd599968e528349ca3cf262120da2831ca8/tools/src/test/resources/2.0/valid-perspective-scopes-2.0.json) groups two text models into one expression scope and describes them as "evaluated as one card."

For the required License mapping, `$..[?(@.type=='machine-learning-model')].licenses`, two fixtures that validate against the pinned #1067 modular schema give the following candidate-policy outcomes:

| Case | Any match within scope | Per-initial-subject presence | Per-initial-subject non-empty |
| --- | ---: | ---: | ---: |
| Two models in one scope; only model A has `licenses` | pass | fail | fail |
| One model with `licenses: []` | pass | pass | fail |

These results compare possible policies; they are not conformance verdicts. Both cases evaluate the initial model subjects directly and do not test attribution after `via`.

For a `required` mapping, is the intended rule (a) at least one matching node within the scope, (b) presence for every initial model selected by the scope, or (c) a consumer-policy choice? Separately, should selecting an empty collection such as `[]` count as present?

If both points are intentionally left to consumer policy, that would also resolve the question. I am not requesting a scoring implementation in this PR; a short clarification or linked follow-up would be sufficient.

Reproduction evidence: [pinned inputs](https://github.com/JunHwan-Kwon/cyclonedx-perspective-semantics-study/blob/evidence-pr1067-7a7d2dd-v3/SOURCE_MANIFEST.json), [F1 before/after trace](https://github.com/JunHwan-Kwon/cyclonedx-perspective-semantics-study/blob/evidence-pr1067-7a7d2dd-v3/results/f1-head-transition-comparison.json), [candidate-policy matrix](https://github.com/JunHwan-Kwon/cyclonedx-perspective-semantics-study/blob/evidence-pr1067-7a7d2dd-v3/results/required-policy-comparison.json), and [schema validation](https://github.com/JunHwan-Kwon/cyclonedx-perspective-semantics-study/blob/evidence-pr1067-7a7d2dd-v3/results/fixture-validation.json).

taleodor:
@JunHwan-Kwon 

Thank you,

**On the F1 trace.** That transition is specified behaviour rather than a regression. `via.referrers` states it directly: "the step replaces the current set with the matching referring objects; objects are not retained merely because they were already in the current set." A risk mapping scoped to model A therefore evaluates `{risk-a}`, not `{model-a, risk-a}`. That is intentional: the mapping selects the risks that affect the model, and the model itself is already selected by the scope.

**On `required`.** The intended reading is (a), with granularity chosen by scoping rather than by the relevance value. The scope is already the unit of evaluation — "tooling generates one view and one completeness evaluation per scope" — so a required mapping is satisfied when it selects at least one node within the scope, and a verdict per model is obtained by declaring one scope per model, as `scope-vision` does. `scope-text` groups both text models deliberately; "Both text models, evaluated as one card" is a request for a single verdict over the pair.

(b) is not well-defined in this model once traversal is in play. After a `via` step the initial subjects are no longer in the evaluation set, by the rule quoted above, so there is nothing to attribute per-subject presence to without back-pointers the specification does not carry. Your matrix notes the same boundary — the fixtures "do not test attribution after `via`". Defining (b) would make `required` mean one thing for direct mappings and something else for traversed ones.

**On empty collections.** Presence is node selection: `licenses: []` is selected, so it is present. Whether an empty collection is *sufficient* is consumer policy. `relevance` and `weight` are importance annotations — "Indicates the importance of a data type to a perspective", and a "relative ranking of mappings" — and this specification deliberately defines no scoring algorithm. An empty-collection rule would be the only conformance rule in a document that otherwise carries none.

I will make this explicit in the text rather than leave it implied, along these lines:

> Relevance and weight rank importance to the audience; they do not define a conformance rule. Within a completeness evaluation, a mapping is satisfied when it selects at least one node within the scope; declare one scope per subject for a verdict per subject. Whether a selected but empty collection is sufficient is left to the consumer's policy.

perspectives/model-card-perspective.json
taleodor: Thank you!

Amended this PR to align:

Verified against #990 head `38dfe9c`, and all three points hold: `bias` is absent from the predefined `impactCategory` enum while `fairness` remains, `bias` is now a weakness kind (`flaw, hazard, bias, predisposing-condition`), and risks associate weaknesses through `relatedWeaknesses` with reusable entries under `definitions.weaknesses`.

One point is slightly stronger than stated: a **custom** impact category is an object (`{"name": ...}`), not a bare string, so `@=='bias'` would not match a custom category either. Against #990 that branch matches nothing at all.

Rather than drop the branch, the mapping now selects the `fairness` **risk domain**, mirroring how Ethical Considerations selects the `ethical` domain:

```
$.risks.risks[?(@.domains[?(@.type=='fairness')])]
```

`fairness` is a predefined risk domain in #990, so the mapping no longer depends on how a methodology categorizes impact, and the two risk mappings are now structurally consistent. The `nativeDescription` is aligned with it.

I have not added a mapping for bias as a weakness. Selecting those would need `relatedWeaknesses` to reach `definitions.weaknesses`, and a traversal step replaces the set of nodes the expression is evaluated against, so it cannot be combined with the existing `referrers` step over `affects`.

JunHwan-Kwon:
@taleodor Thank you. This resolves my completeness question.

I reran the bounded License probes at `d3fca0c` and recorded the clarified interpretation as any-match per scope, with one scope per subject when a per-subject verdict is needed. Selecting `licenses: []` establishes node presence, while whether that value is sufficient remains consumer policy.

I will retain F2 and F5 as non-normative policy comparisons, with the intended interpretation now recorded, and consider this question resolved.

Updated evidence: [pinned inputs](https://github.com/JunHwan-Kwon/cyclonedx-perspective-semantics-study/blob/evidence-pr1067-d3fca0c-v1/SOURCE_MANIFEST.json) and [required-policy replay](https://github.com/JunHwan-Kwon/cyclonedx-perspective-semantics-study/blob/evidence-pr1067-d3fca0c-v1/results/required-policy-comparison.json).

perspectives/model-card-perspective.json
JunHwan-Kwon: Thank you. The domain-based expression addresses the cross-PR vocabulary alignment I raised.

I replayed the existing fixture at `d3fca0c`. The `referrers` step still produces `{risk-a}` for model A, while Fairness Assessments selects no node because the unchanged fixture declares only the `ethical` domain. The pinned #1067 risk schema does not yet include `fairness` as a predefined risk domain, while it is present at #990 head `38dfe9c`.

I have recorded this as a cross-PR schema and fixture dependency, not as a defect in the revised expression. A positive fairness-domain case can be checked after the corresponding risk-model integration. I have no further change to request on this expression.

Reproduction evidence: [fairness-domain integration observation](https://github.com/JunHwan-Kwon/cyclonedx-perspective-semantics-study/blob/evidence-pr1067-d3fca0c-v1/results/fairness-domain-integration-observation.json).

perspectives/model-card-perspective.json
mrutkows: I see in https://github.com/relizaio/cyclonedx-specification/pull/2 :

```
"predefined": "cdx:perspectives:model-card",
```
Does this need to change here since that registry  PR was merged?

perspectives/model-card-perspective.json
mrutkows: The name could be fine, but I have been viewing the "name" and "version" (and perhaps more loosely "description") as alternative "model identifiers.  This helps identify models that are not published necessarily to a model repos./registry.  This would help me perhaps in creating test builds, model checkpoint builds, etc. Both name and description are "baked in" to thingslike purls, but even some down-stream repos. like Ollama have no "id"; within its registry domain, it has an "org.", "name" and "version" which has its own format (with names like "granite4.2:latest" or "granite4.2:3b-q3_K_L")

perspectives/model-card-perspective.json
```diff
@@ -0,0 +1,248 @@
...
+          "expression": "$..[?(@.type=='machine-learning-model')].identifiers",
+          "nativeName": "Model Identifiers",
+          "nativeDescription": "Identifiers under which the model is published or catalogued, such as a Package-URL for a model hub entry, grouped by the party asserting them.",
+          "relevance": "recommended",
+          "weight": 0.7,
+          "rationale": "Name and version alone are ambiguous across hubs and forks; asserted identifiers let a card be matched to the published artefact."
+        },
+        {
+          "expression": "$..[?(@.type=='machine-learning-model')].parties[?(@.roles[?(@.role=='supplier')])]",
+          "nativeName": "Developed By",
```
mrutkows, line 39: Assume we can have a "Published At" as a means to express the model repo. (e.g., Hugging Face, Ollama, LM Studio, etc.)

perspectives/model-card-perspective.json
```diff
@@ -0,0 +1,248 @@
...
+        },
+        {
+          "expression": "$.formulation[*]",
+          "via": [
+            {
+              "refs": "$..[?(@.type=='machine-learning-model')].modelProperties.training.formula"
+            }
+          ],
+          "nativeName": "Training Procedure",
+          "nativeDescription": "The formula describing how the model was trained: its workflows and tasks, such as pre-training, fine-tuning, and alignment stages, and the data each consumed.",
```
mrutkows, line 109: We can use CycloneDX semantics for "native" names.  Although I was not a fan of "formula" I would like CDX perspective to name things according to our 2.0 element/object names (e.g., "Training formula", "Fine-tuning formula", etc.)

schema/2.0/model/cyclonedx-definition-2.0.schema.json
```diff
@@ -25,6 +25,15 @@
...
+          "type": "array",
+          "uniqueItems": true,
+          "title": "Perspectives",
+          "description": "Reusable perspective declarations. Each entry is a complete inline perspective, identified by its bom-ref, that may be applied from the document's perspectives by reference, optionally scoped to different parts of the document at each application. A definition declares no scopes of its own.",
+          "items": {
+            "$ref": "cyclonedx-perspective-2.0.schema.json#/$defs/perspectiveDefinition"
+          }
         }
       }
     }
```
mrutkows, line 39: Before closure, want to confirm we can extract all properties on a model component that match a certain panel.  My example/use case is the 1.7 card I gen'ed for our 4.2 models:

```
 "properties": [
        { "name": "cdx:ai-ml:model:modality:text", "value": "true" },
        { "name": "cdx:ai-ml:model:modality:code", "value": "true" },
        { "name": "cdx:ai-ml:model:parameter:count", "value": "8B" },
        { "name": "cdx:ai-ml:model:parameter:tune_method", "value": "full" },
        { "name": "cdx:ai-ml:model:hyperparameter:num_hidden_layers", "value": "40" },
        { "name": "cdx:ai-ml:model:hyperparameter:hidden_size", "value": "4096" },
        { "name": "cdx:ai-ml:model:hyperparameter:intermediate_size", "value": "12800" },
        { "name": "cdx:ai-ml:model:hyperparameter:num_attention_heads", "value": "32" },
        { "name": "cdx:ai-ml:model:hyperparameter:num_key_value_heads", "value": "8" },
        { "name": "cdx:ai-ml:model:hyperparameter:hidden_act", "value": "silu" },
        { "name": "cdx:ai-ml:model:hyperparameter:layer_norm_type", "value": "RMSNorm" },
        { "name": "cdx:ai-ml:model:hyperparameter:context_length", "value": "131072" },
        { "name": "cdx:ai-ml:model:hyperparameter:max_position_embeddings", "value": "131072" },
        { "name": "cdx:ai-ml:model:hyperparameter:vocab_size", "value": "49152" },
        { "name": "cdx:ai-ml:model:hyperparameter:quantization", "value": "bf16" },
        { "name": "cdx:ai-ml:model:hyperparameter:tokenizer_class", "value": "PreTrainedTokenizerFast" },
        { "name": "cdx:ai-ml:model:hyperparameter:_undefined:attention_head_dim", "value": "128" },
        { "name": "cdx:ai-ml:model:hyperparameter:_undefined:positional_embeddings", "value": "RoPE" },
        { "name": "cdx:ai-ml:model:hyperparameter:_undefined:rope_theta", "value": "10000000" },
        { "name": "cdx:ai-ml:model:hyperparameter:_undefined:transformer_class", "value": "GraniteForCausalLM" },
        { "name": "cdx:ai-ml:model:hyperparameter:_undefined:architecture_type", "value": "dense-decoder-only" },
        { "name": "cdx:ai-ml:model:hyperparameter:_undefined:attention_mechanism", "value": "GQA" },
        { "name": "cdx:ai-ml:model:hyperparameter:_undefined:shared_input_output_embeddings", "value": "false" },
        { "name": "cdx:ai-ml:model:hyperparameter:_undefined:long_context_extension_method", "value": "staged-512K-RoPE-scaling" },
        { "name": "cdx:ai-ml:model:language", "value": "en,de,es,fr,ja,pt,ar,cs,it,ko,nl,zh" },
        { "name": "cdx:ai-ml:model:architecture:topology", "value": "dense" },
        { "name": "cdx:ai-ml:model:architecture:secondary", "value": "rotary-position-embedding" },
        { "name": "ibm:ai-ml:model:disclaimer", "value": "Structured dataset metadata used for pre- and post-training IBM Granite 4.2 base and instruct models will be released as part of IBM's effort to provide the community with highly transparent open-source model development, pending completion of internal review. This structured disclosure will be provided in machine-readable JSON format as a governance artifact to inform risk-based controls for downstream model use." },
        { "name": "ibm:ai-ml:model:architecture:transformer_class", "value": "GraniteForCausalLM" },
        { "name": "ibm:ai-ml:model:hyperparameter:attention_head_dim", "value": "128" },
        { "name": "ibm:ai-ml:model:compliance:iso42001", "value": "certified" }
      ]
```
and will have this for other model "components" such as the tokenizer:
```
          "properties": [
            { "name": "cdx:ai-ml:model:tokenizer", "value": "PreTrainedTokenizerFast" },
            { "name": "cdx:ai-ml:tokenizer:hyperparameter:vocab_size", "value": "49152" },
            { "name": "cdx:ai-ml:tokenizer:hyperparameter:bos_token", "value": "<|startoftext|>" },
            { "name": "cdx:ai-ml:tokenizer:hyperparameter:eos_token", "value": "<|endoftext|>" },
            { "name": "cdx:ai-ml:tokenizer:hyperparameter:pad_token", "value": "<|pad|>" }
          ]
```

of course, for the CDX perspective want only the "cdx:" prefix names... whenever a component appears (i.e., a "component" object should pull all its CDX properties)

mrutkows (changes requested):
Looks good and thanks for your work; it is very, very important. 

Could you help me see the previous "ask" where I wanted a means to request that any appearance of "bom-ref" (by default or within specified objects) would cause those referenced objects to be pulled into the perspective output document.

perspectives/model-card-perspective.json
taleodor: Shouldn't change. This is an arbitrary internal ref within perspective json, it doesn't represent perspective's published identity (and mustn't represent it). The identity lives in the registry schema/perspectives-defs.json which maps to file by naming convention.

perspectives/model-card-perspective.json
taleodor: We can split those into different mappings, lmk if you want that.

perspectives/model-card-perspective.json
taleodor: We can add a separate mapping, or modify this one, lmk how you want it + can discuss more on call.

perspectives/model-card-perspective.json
taleodor: Again, can be renamed or split to different mappings. We can discuss.

schema/2.0/model/cyclonedx-definition-2.0.schema.json
taleodor: This should all be expressible but we would need to add mappings accordingly if we want this to be part of predefined model-card.

taleodor:
@mrutkows 

> Could you help me see the previous "ask" where I wanted a means to request that any appearance of "bom-ref" (by default or within specified objects) would cause those referenced objects to be pulled into the perspective output document.

## 1. Scope: referenced objects are pulled in automatically

A scope selects the part of the document the perspective applies to, and it can be declared **by reference** or **by path**:

```jsonc
"scopes": [
  {                                   // by ref — bom-ref / bom-link
    "bom-ref": "scope-vision",
    "name": "Vision model",
    "targets": [ "model-vision" ]
  },
  {                                   // by path — RFC 9535 JSONPath from the document root
    "bom-ref": "scope-text",
    "name": "Text models",
    "description": "Both text models, evaluated as one card.",
    "expressions": [
      "$.components[?(@.type=='machine-learning-model' && @.group=='text')]"
    ]
  }
]
```

## 2. Mappings: `via` steps, when you want to state the traversal explicitly

Where the data is *related to* the scope rather than contained in it, a mapping declares traversal steps. Three kinds, applied in order, each to the result of the previous:

```jsonc
// refs — resolve reference values found at a path
{
  "expression": "$..[?(@.type=='data')]['name','version','identifiers','licenses']",
  "via": [ { "refs": "$..[?(@.type=='machine-learning-model')].modelProperties.training.datasets[*]" } ],
  "nativeName": "Dataset Identity and Licensing"
}

// referrers — the inverse: objects that hold a reference to the current nodes
{
  "expression": "$.risks.risks[?(@.domains[?(@.type=='ethical')])]",
  "via": [ { "referrers": "$.risks.risks[*].affects[*]" } ],
  "nativeName": "Ethical Considerations"
}

// dependencies — walk the dependency graph, optionally transitively
{
  "expression": "$.components[?(@.type=='cryptographic-asset')]",
  "via": [ { "dependencies": "depends-on", "transitive": true },
           { "dependencies": "provides" } ],
  "nativeName": "Algorithms in Use"
}
```

## 3. Possible gaps

**1. The reference-following matches only one of the two shapes CycloneDX uses.**
A reference is followed when it appears as an object carrying a `ref` property. Most relationships in 2.0 are encoded as bare reference strings instead. `training.datasets[]` resolves to `refLinkType`, and `dependsOn`, `provides` and `affects` are arrays of plain values. Those are skipped. (Can still be expressed by other means).

**2. Only outbound references are followed.**
References contained in the scoped subtree are followed. Objects that point *at* the scoped node from elsewhere are not reached, because they sit outside that subtree. This matters most for `dependencies`, which is a document-root array: an entry such as `{"ref": "model-vision", "dependsOn": ["lib-openssl"]}` is a sibling of `components`, not part of the model's subtree, so scoping a model never pulls in its dependency graph at any depth. The same applies to `risks[].affects[]`. Reaching either requires a mapping with `via.referrers` or `via.dependencies`.

**3. Following stops after one hop.**
A reference is resolved and its target selected, but that target's own references are not followed in turn. If A references B and B references C, the scope contains B and never C; C has to be enumerated in the scope or reached with a `via` step.

**4. There is no control over which references are followed.**
Following is unconditional: every qualifying reference inside the scope is followed, always. There is no way to express "follow the training-dataset references but not the rest".

**5. A `via` step replaces the working set rather than adding to it.**
Each step produces the nodes it reaches, and nodes are not retained merely because they were already in the set. A single mapping therefore cannot express "the scoped nodes *and* what they reference". It is one or the other.

schema/2.0/model/cyclonedx-definition-2.0.schema.json
```diff
@@ -25,6 +25,15 @@
         },
         "businessObjectives": {
           "$ref": "cyclonedx-business-objective-2.0.schema.json#/$defs/businessObjectives"
+        },
+        "perspectives": {
```
stevespringett, line 29: Perspectives are a root element. There should not be a need to additionally define them here.

stevespringett (changes requested):

jkowalleck:
RFC notice sent on September 16, 2026.
Public RFC period ends October 14, 2026.

PR #1068, opened 2026-08-31
linter: add check ref-best-practice

jkowalleck:
this one transforms an existing semantics check into a linter check


jkowalleck:
tests passing: https://github.com/CycloneDX/specification/actions/runs/33398704467/job/99509486280?pr=1068

PR #1075, opened 2026-09-02
Inventory refactor and definitions registry

### LINKED ISSUE #1072: [2.0]: Definitions registry and multi-product inventory refactor
## Objective

CycloneDX 2.0 restructures how documents declare *what exists* versus *what is asserted*. Reusable objects, including components, parties, licenses, profiles, release notes, and certifications, can be declared once in `/definitions` and referenced from anywhere in the document. A reference always denotes an identical object; anything that varies by usage is declared inline where it is used.

Product inventory becomes an explicit, self-contained statement: each entry names its subject and lists what that subject contains, so a single document can describe the contents of one product or many, with shared components declared once.

## Why this refactor is necessary

Every other assertion in the specification is self-contained and names its own subject: declarations name their targets, vulnerabilities name what they affect, compositions name their assemblies, and threats name their scope. Inventory is the one assertion that instead relies on an implicit convention, where the root `components` array represents the inventory of `metadata.component` by rule rather than by reference.

That convention is a leftover from when CycloneDX was a BOM-only format, which has not been true since v1.4. A CycloneDX document may be a VEX, an attestation, a threat model, or any combination of capabilities, yet inventory declaration still assumes one document, one BOM, one subject. This refactor brings inventory in line with how everything else in the specification already works, and removes the single-subject assumption in the process.

Goals:

- One way to declare a product's contents, whether the document covers one product or fifty.
- Eliminate duplication of repeated objects such as licenses, suppliers, and shared components.
- Consistent reference behavior across the specification, applied by the same rules in future versions.

## Related issues addressed in whole or in part

- [ ] #734 Inconsistent use of mediaType, mimeType, and contentType
- [ ] #263 Dedicated node for licenses

Part of the CycloneDX 2.0 effort tracked in #631

stevespringett:
Implements the inventory and definitions refactor for CycloneDX 2.0 (closes #1072).

Product inventory becomes an explicit, self-contained statement: each `inventories` entry names its subject and lists what it contains, replacing the implicit `metadata.component` + root `components` convention. Reusable objects (components, parties, licenses, profiles, release notes, certifications) can be declared once in `/definitions` and referenced from anywhere; a reference always denotes an identical object, and anything that varies by position is declared inline.

Some minor other changes and corrections along the way.


jkowalleck:
before detailed reviewing, could you fix those merge conflicts?

jkowalleck:
RFC notice sent on September 02, 2026

* https://groups.io/g/CycloneDX/topic/121045570
* https://cyclonedx.slack.com/archives/CVA0G10FN/p1788316773546949

Public RFC period ends September 30, 2026

JunHwan-Kwon:
@stevespringett @taleodor At #1075 head `24b8d8e`, I reproduced a bounded regression in the existing perspective fixture.

In `tools/src/test/resources/2.0/valid-perspective-2.0.json`, both sample components are now under `definitions.components`, while the four component-related mappings still begin with `$.components[*]`. Evaluating all seven JSONPath expressions with an RFC 9535 evaluator gives match counts of `0, 0, 0, 0, 1, 1, 1` in mapping order, although all seven mappings are marked `relevance: "required"`.

The history separates the new regression from two pre-existing stale mappings. The inventory refactor in `be91401` caused the name and version mappings to drop from two matches each to zero. The identifier mapping had already become stale at `10b343a`, when direct `purl` and `cpe` values were removed from the fixture; the supplier mapping became stale at `caf17eb`, when supplier moved to role-bearing `parties`. Retargeting only the root prefix from `$.components` to `$.definitions.components` therefore restores name and version but leaves supplier and identifier at zero.

The fixture remains JSON Schema-valid. The checked-in Java, JavaScript, and PHP 2.0 fixture-validation paths perform JSON Schema validation, but do not execute these perspective mappings or assert match cardinality. Could the fixture and mappings be migrated together in #1075, with a small semantic mapping smoke test, or should that work be tracked in a linked perspective-mapping issue?

jkowalleck:
@stevespringett , could you fix tose merge conflicts?

PR #1080, opened 2026-09-03
2.0 dev fix links

jkowalleck:
the resulting URLs are properly picked up by the docs tool and rendered as actual clickable URLs.


jkowalleck:
pure documentational change that improves the docs rendering

PR #1113, opened 2026-09-17
Changed bom-ref to refId. Title and description updated. Updated test…

### LINKED ISSUE #673: Consider changing bom-ref to bom-id in 2.0
### Discussed in https://github.com/CycloneDX/specification/discussions/671

<div type='discussions-op-text'>

<sup>Originally posted by **andreas-hilti** August 10, 2025</sup>
I'm wondering about the identifier "bom-ref": shouldn't this rather be called "bom-id"?

If you look at the description https://cyclonedx.org/docs/1.6/json/#components_items_bom-ref
> An optional identifier which can be used to reference the component elsewhere in the BOM.

For me, the component has an identifier (which I'd rather call "bom-id"), and this bom-id is then referenced in other places, e.g. in dependencies https://cyclonedx.org/docs/1.6/json/#dependencies_items_ref
> References a component or service by its bom-ref attribute

which should rather be:
> References a component or service by its bom identifier attribute (bom-id)

Only where it is used/referenced, this is actually a reference.

Could this be considered for CycloneDX 2.0?
For me, the main benefit would be clarity and alignment with standard usage of the term "reference".</div>

stevespringett:
- Closes #673


jkowalleck:
We might consider renaming `cyclonedx-common-2.0.schema.json#/$defs/refType` to `cyclonedx-common-2.0.schema.json#/$defs/refIdType`

jkowalleck:
RFC notice sent on September 18, 2026

Public RFC period ends October 16, 2026

schema/2.0/model/cyclonedx-behavior-2.0.schema.json
```diff
@@ -55,7 +55,7 @@
         "actors": {
           "type": "array",
           "title": "Actors",
-          "description": "References using bom-link or bom-ref to objects that perform this behavior.",
+          "description": "References using bom-link or refId to objects that perform this behavior.",
```
jkowalleck, line 58: Maybe out of scope but a good situation to do this: in the docs: make sure all `ref` are marked a code.

```suggestion
          "description": "References using bom-link or `refId` to objects that perform this behavior.",
```


Ir if not in scope, create a follow-up ticket from this comment.

jkowalleck, line 58: Maybe out of scope but a good situation to do this: 
make sure all `BOM-Link` are stylized correctly.

```suggestion
          "description": "References using BOM-link or refId to objects that perform this behavior.",
```

Ir if not in scope, create a follow-up ticket from this comment.

schema/2.0/model/cyclonedx-common-2.0.schema.json
jkowalleck: Please consider renaming `cyclonedx-common-2.0.schema.json#/$defs/refType`
to `cyclonedx-common-2.0.schema.json#/$defs/refIdType`


I think this is the right scope for this, since the `refType` is the root of all the `refId`, and should follow the new naming schema.

jkowalleck (approved):

oej:
Good change. Approved. /O

PR #1114, opened 2026-09-18
ci: zizmor fix ref to reusable workflows

jkowalleck:


jkowalleck:
ci passing: https://github.com/CycloneDX/specification/actions/runs/35312742318/job/105497938982?pr=1114

